PR #11, opened 2016-12-13
Updates to type classes & minor edits

LiamGoodacre:
There's still more to do but I think this is better than what we had.  Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,10 @@ PureScript is a small strongly, statically typed compile-to-JS language with a n
 
 - Type Inference
 - Higher Kinded Polymorphism
-- Support for basic Javascript types
+- Support for basic JavaScript types
 - Extensible records
 - Extensible effects
-- Optimizer rules for generation of efficient Javascript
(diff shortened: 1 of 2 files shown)


hdgarrood:
Sorry, but I don't think this is the style of documentation we want here, really. The stuff inside the language/ directory is really meant to be a reference, i.e. aimed at people who have some idea of how this part of the language works already &mdash; there's already PureScript by Example and plenty of Haskell resources for at people who haven't come across typeclasses before. I think the documentation here should be more direct and concise.

Also I don't think we want to link to 2014's 24 days of PureScript, it's probably too far out of date now.

LiamGoodacre:
Thanks for your feedback, that makes sense.
I've trimmed my changes down to just minor edits :)

hdgarrood:
That looks good :) One tiny nit - can you please use the full version of the bower command i.e. `bower install --save` rather than `bower i -S`?

hdgarrood:
Thanks!

PR #12, opened 2016-12-15
Fix operators and add a link to `Data.Int.Bits`

seungha-kim:


language/Values.md
```diff
@@ -175,7 +175,7 @@ Function   | JS Operator | Meaning
 `-`        | `-`         | Numeric subtraction
 `*`        | `*`         | Numeric multiplication
 `/`        | `/`         | Numeric division
-`%`        | `%`         | Numeric modulus
```
garyb: This isn't quite right actually - `mod` does not behave like `%` in JS, since it returns zero for floats. `purescript-math` does provide JS's `%` operator though.

chexxor:
LGTM

paf31:
Thanks!

PR #16, opened 2017-01-11
Add "Indentation in binding blocks"

chexxor:


chexxor:
Interestingly, if `let` is 4 spaces, its following `in` can be 2 spaces.
Same for `in` block - if `in` is 2 spaces, its body can be indented 1 space.
Same for `do` block - if `do` is 6 spaces, its body can be indented 1 space.
These are all absolute indentation, where absolute means from the beginning of the line, not the parent's indentation level.

splodingsocks:
Awesome, thanks for adding that, @chexxor !

hdgarrood:
This is useful information to have but it should be in the Syntax.md file instead of this one.

hdgarrood:
Hmm, it seems like much of what's currently in the 'Values' file is in the wrong place, actually. I'll merge this for now and we can address that later.

PR #21, opened 2017-01-15
Move Values.md into Syntax.md, fixes #17

### LINKED ISSUE #17: Move parts of Values page into Syntax page
Some information in the Values page should be in the Syntax page. See https://github.com/purescript/documentation/pull/16#issuecomment-271960513

hdgarrood:
* Removed table showing correspondence between JS operators and Prelude
  operators, as it suggests that these operators are built in to the
  language.
* Fix misleading section heading 'user-defined operators' for the same
  reason
* Moved the rest of language/Values.md into language/Syntax.md as
  everything that was left was really about syntax.
* Update operator definition example to be more self-contained.


hdgarrood:
This raises the question of where that operator table should go. I'm fairly confident that the language reference is not the right place for it. Other options: a guide in `guides/`, or alongside the `Prelude` documentation.

garyb:
I think a guide would make sense for that kind of thing 👍 since it also needs to refer to `Math` to deal with `%` , since `mod` is not that.

hdgarrood:
Ok cool, I'll push another commit or two in a bit.

hdgarrood:
I think this is ready for review now.

LiamGoodacre:
Oooh I didn't know you could do ``` ``foo `bar` baz`` ``` to get ``foo `bar` baz``. Nice! 😄

hdgarrood:
Nor did I until about 20 minutes ago!

guides/Common-Operators.md
```diff
@@ -42,18 +50,46 @@ This means that the PureScript analogues to unary JavaScript operators (other th
...
+after dividing by 3 is 1, and indeed in JS, `10 % 3 == 1`. Likewise,
+in PureScript, ``10 `mod` 3 == 1``.
+
+It's a bit harder to say what a 'remainder' operation should mean if
+we are trying to extend it to real (non-integral) numbers,
+though. Real numbers can be divided into one another with nothing
+'left over', i.e. in some sense, the remainder of dividing one number
+by another is always 0.
+
+Since JS just has one number type, it has to come up with a sensible
```
chexxor: What's the reason for splitting one long line into multiple lines? I would guess having one long line will allow browser to better reflow text.

hdgarrood: Browsers treat newlines the same as any other space unless you have a `white-space: pre` CSS rule (which this won't). I used hard line breaks because I think it's easier to read/deal with the text in the source Markdown format this way (although I can change it back if people don't agree).

chexxor: I think it's fine if there's a good reason for it.

I usually don't use hard line breaks when writing prose because it's painful to manually reflow the lines if I decide to revise a paragraph after it's written. It's a relatively large mental process that doesn't produce have much value.

It looks like Github's in-browser editor soft-wraps long lines, so long lines shouldn't be a problem there. In which editor is it a problem? Or perhaps you don't mind manually reflowing lines?

hdgarrood: > it's painful to manually reflow the lines if I decide to revise a paragraph after it's written

Not in Vim or Emacs (I expect other editors can do this too, but I haven't used other editors much so I can't say for sure). I only consider using hard breaks at all because it's very easy to reflow paragraphs or an entire file with a few keystrokes.

chexxor: I didn't know vim could do that - pretty neat. Soft breaks seems like a more universally standard algorithm than an editor's macro for doing it. I'll need to look at other projects to see how they're doing it before I can satisfyingly settle on it, though. Should we apply the same formatting to all files in this project, then?

hdgarrood: Come to think of it, it could put people off sending a simple PR inside GitHub's web interface, if they think they might be asked to reflow text. I've since learned how to make editing soft-wrapped text comfortable in emacs anyway. I think I'll change it back.

Btw now that GitHub can do 'rich' diffs of markdown files, I think either way is fine wrt diffing.

chexxor:
Thanks for organizing this, @hdgarrood! Much appreciated!

PR #27, opened 2017-01-17
General reorganization, delete out-of-date pages.

### LINKED ISSUE #3: Remove pages
I think we can remove quite a few pages:

**Out of date/no longer necessary**

- 0.7/0.10 Migration Guide
- PureScript Conf 2015/2016
- Recommended Libraries (very out of date now, and no longer necessary now that we have the package set repo and Pursuit?)
- PureScript and Docker (not sure if this still works, but we should check)

**Belong elsewhere**

- PSCi (should be in the main repo as a README)
- Frequently asked `psc-ide` questions (same)
- How to implement a foreign function with a constraint (merge with FFI.md?)
- The Partial Type Class (on the website now - I'm fine with articles living here eventually, but let's pick one?)
- Documentation Admin (let's make issues for this)
- Wikipedia Page Proposal (this is still good, but as a draft, I'm not sure it belongs here)

paf31:
This moves a few files around, adds a few headers, and deletes the pages mentioned in #3.

Fixes #3.

After this, I'll probably start working on the error message pages, since most of those need quite a bit of work.


paf31:
It might be easier to [navigate this using the file explorer](https://github.com/purescript/documentation/tree/phil/reorg) than to look at the diff.

hdgarrood:
I think I'd prefer to have the Partial type class guide here than on the website, actually. Same for the other pages under `learn/` on the website, too - just so that we don't have a situation where it's hard to work out where to put things because some things are here and others are on the website.

hdgarrood:
Other than that, :+1:

paf31:
The problem is that lots of links to those pages already exist. But I agree that one place is better, so maybe the website should link here?

cc @garyb

hdgarrood:
Good point. Yes, I think that would be best (afaik we can't do real redirects on GH pages, sadly).

garyb:
Works for me 👍 

> afaik we can't do real redirects on GH pages, sadly

We could at least put JS redirecting / "go here" style pages up for the existing pages on the site when we switch over.

paf31:
I've copied over the missing articles and put back the ones I deleted. We still need to update http://purescript.org/learn, but I'll merge this for now.

Issue #23, opened 2017-01-16
Update Partial guide to mention unsafePartialBecause


hdgarrood:
This function is a really good idea and currently it's only mentioned in the generated API docs for `purescript-partial`, AFAIK.

hdgarrood:
See #33

PR #33, opened 2017-01-28
Recommend using `unsafePartialBecause`

chexxor:
Fixes https://github.com/purescript/documentation/issues/23

### DIFF
--- a/guides/The-Partial-type-class.md
+++ b/guides/The-Partial-type-class.md
@@ -58,8 +58,8 @@ at src/Main.purs line 8, column 1 - line 8, column 56
 
 *Aside: Yes, this is not a fantastic error. It's going to get better soon.*
 
-The solution is usually to add an application of `unsafePartial` somewhere,
-like this:
+The solution is usually to add an application of `unsafePartial`,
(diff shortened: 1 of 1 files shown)


hdgarrood:
Looks great! It might be worth adding that `unsafePartial` should probably be preferred if you're worried about performance, since `unsafePartial` will be inlined as of the next compiler release.

chexxor:
Great idea @hdgarrood! I've updated the branch to include a commit which includes that inlining bit.

hdgarrood:
I'm sorry to mess you around, but actually I've been thinking about `unsafePartialBecause` and it's actually not so clear to me that we should be recommending it. I'm going to open an issue on purescript-partial when I get a moment.

chexxor:
Sure, no problem.

hdgarrood:
https://github.com/purescript/purescript-partial/issues/7

hdgarrood:
Seems like we shouldn't be recommending `unsafePartialBecause` after all (see discussion on purescript/purescript-partial#7), so I'm going to close this. Thanks though!

PR #35, opened 2017-01-29
Move missing pages into errors/

paf31:
I figured it's better that the compiler link to _some_ page than a 404 page, so this moves the pages under `errors/missing/` into `errors/`.


paf31:
Actually, I wonder if symbolic links would work better. Then at least people can see the `missing` directory in GitHub.

hdgarrood:
Perhaps we could also go ahead and move them all in, and at the same time create an issue with a checklist of all the ones that are currently in `missing`? That way it's easy to see how many errors/warnings still need to have something written about them.

paf31:
Ok sounds good.

paf31:
I've created #36, so I'll go ahead and merge this.

PR #40, opened 2017-02-06
TypesDoNotUnify - Arrays

kelden:
Added a section on arrays with examples from the [PureScript by Example book](https://leanpub.com/purescript/read#leanpub-auto-simple-types). I'm not sure if this section is necessary but it might save someone a headache in the future.

### DIFF
--- a/errors/TypesDoNotUnify.md
+++ b/errors/TypesDoNotUnify.md
@@ -35,7 +35,7 @@ In the example above, the types `String` and `Int` can never be made equal, henc
 ## Notes
 ### Arrays
 
-Unlike in Javascript, all elements of a PureScript array must be the same type. For example, ```[true, false``` or ```[1, 2, 3]``` work fine but ```[1, false]``` does not.
+Unlike in Javascript, all elements of a PureScript array must be the same type. For example, ```[true, false]``` or ```[1, 2, 3]``` work fine but ```[1, false]``` does not.
 
 ### Another Example
 
--- a/errors/TypesDoNotUnify.md
+++ b/errors/TypesDoNotUnify.md
@@ -35,7 +35,7 @@ In the example above, the types `String` and `Int` can never be made equal, henc
 ## Notes
 ### Arrays
 
-Unlike in Javascript, all elements of a PureScript array must be the same type. For example, ```[true, false]``` or ```[1, 2, 3]``` work fine but ```[1, false]``` does not.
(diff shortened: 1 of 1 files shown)


errors/TypesDoNotUnify.md
```diff
@@ -33,6 +33,9 @@ In the example above, the types `String` and `Int` can never be made equal, henc
 - Look carefully at the error, especially the information at the end. Usually, it will help to narrow down the offending expression. For example, in the error message above, we are told that the error occurred "while checking that expression `"test"` has type `Int`".
 
 ## Notes
+### Arrays
+
+Unlike in Javascript, all elements of a PureScript array must be the same type. For example, ```[true, false]``` or ```[1, 2, 3]``` work fine but ```[1, false]``` does not.
```
LiamGoodacre: Nit-pick: JavaScript has a capital S

paf31: We should also mention here that this will fail because the compiler uses unification to verify this property.

chexxor: How would you explain type unification here? Something like: "This fails because the compiler attempts to unify this type as follows: The outer-most type is `Array a`, so the type-checker finds `a` by using the type of its contents. It finds the contents' type by checking the type of each elements, and reducing the results to a single value. It happens to read the type of the last element first, which is `Boolean` for `false`, using that as `a`, then verifying all elements are `Boolean`. This type-check fails when reading `Int` for `1`, so throws the error:

```
  Could not match type
           
    Boolean
           
  with type
       
    Int
```

paf31: I would just say that the compiler unifies the inferred types of the array elements to make sure they're the same. Although, there are lots of uses of unification in the compiler, and I don't know if we need to give an example for each one.

kelden: @LiamGoodacre Thanks!

@paf31 I don't think I know Haskell/Purescript internals enough to adequately illustrate the use of unification here or why it is necessary.

paf31: @kelden There are lots of uses of unification, and I don't think we need a subsection for each one. So I would actually just phrase this as an additional example, with a one sentence explanation of the error ("this error occurs because PureScript arrays must have elements with the same types, and the compiler failed to unify the inferred types of the array elements").

chexxor:
LGTM

paf31:
Thanks!

PR #44, opened 2017-02-19
Remove co-learning meetup

chexxor:
Turnout is low and I think IRC and Slack are better, more convenient venues for newcomers.


paf31:
👍 Thanks

PR #45, opened 2017-02-20
Add example and explaination of InfiniteType error

dgendill:
I'm not entirely sure my explanation of the error is correct, but the example and suggested solution might be helpful for anyone else who runs into this error.

### DIFF
--- a/errors/InfiniteType.md
+++ b/errors/InfiniteType.md
@@ -3,50 +3,11 @@
 ## Example
 
 ```purescript
-module InfiniteTypeErrorExample where
+module ErrorExample where
 
-import Control.Coroutine as CR
-import Control.Coroutine.Aff as CRA
-import Control.Monad.Aff (Aff, launchAff)
-import Control.Monad.Aff.AVar (AVAR)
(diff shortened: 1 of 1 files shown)


paf31:
Thanks for putting this together, but I think we should try to find a minimal example. Here's a suggestion:

```purescript
applySelf x = x x
```

This fails because `x` must have a function type, but the domain of that function must be the same type, so we end up with an infinite type `(((... -> r) -> r) -> r) -> r`

dgendill:
Ah, ok, that makes more sense now.  Yeah, there's a lot of extra code in my example - I'm still working through why it throws that error and why changing the types fixes it, but your example helps me see what the underlying problem is.  I went ahead and just put your example in the documentation, but I'm not sure what level of detail and explanation we're looking for, so if that's not what we're after feel free to close this PR.

hdgarrood:
This is a good start, I think; certainly better than what we have now. Could you please use backticks `` ` `` around the code parts so it renders like in @paf31's comment, and also put a full stop at the end?

paf31:
Thanks!

PR #47, opened 2017-02-27
Update EscapedSkolem.md

coot:


errors/EscapedSkolem.md
```diff
@@ -26,6 +26,10 @@ In the example above, the type of `runST` causes us to check `newSTRef 0` agains
 
 One common pitfall is to use the `runST` or `runPure` functions with the `$` operator. This will often lead to the `EscapedSkolem` error due to instantiation of type variables. Avoid this by using parentheses instead of `$`.
 
+### Pointfree style
+
+In some cases point free style can also leak type variables.
```
paf31: Do you have a short example to go with this?

Also, nitpick: you alternately use "pointfree" and "point free" here.

coot: At the moment I don't but I will try to think about a simple one.

Issue #22, opened 2017-01-15
Overlap between language/Syntax.md and language/Records.md


hdgarrood:
Record literals and record updates are documented twice, for instance. I think we could remove the entire `language/Records.md` file.

PR #48, opened 2017-03-01
Document syntax of case-expressions.

chexxor:
I came to lookup how to delimit multiple values of a case-expression but didn't find it in the Syntax doc. It *is* in the [Pattern-Matching doc](https://github.com/purescript/documentation/blob/master/language/Pattern-Matching.md), however.

What are the rules for organizing language directory? I saw "Syntax" and figured case expressions would be in there because it's special syntax. Because everything in the language directory should be about syntax or core principles, should we reorganize this by deleting this file and putting it into other doc files, one for each section?

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -392,6 +392,19 @@ Operator sections also work for functions used this way:
 fooBy2 = (_ `foo` 2)
 ```
 
+## Case expressions
+
+The `case` and `of` keywords are used to deconstruct values to create logic based on the value's constructors. You can match on multiple values by delimiting them with `,` in the head and cases.
+
+``` purescript
+f :: Maybe Boolean -> Either Boolean Boolean -> String
+f a b = case a, b of
+  Just true, Right true -> "Both true"
+  Just true, Left _ -> "Just is true"
+  Nothing, Right true -> "Right is true"
+  _, _ -> "Both are false"
+f (Just true) (Right true)
+```
 
 ## If-Then-Else expressions
 


hdgarrood:
The "Pattern Matching" page isn't very long, and I agree with what you thought — in my mind, pattern matching is part of the syntax — so maybe it would make sense to fold the "Pattern Matching" page into "Syntax".

With respect to what you've written in this PR, do you think it might be good to about first introduce the simplest case of a `case` expression which only matches one expression, and then separately mention that you can use commas to match on multiple expressions, with a separate example?

chexxor:
@hdgarrood, are you worried about the Syntax page getting too big? I'm a bit worried about that because the "syntax" category can become really large. We could break it up into smaller pages doc pages. This suggestion might in contrast to your previous thoughts on this topic, it seems: https://github.com/purescript/documentation/issues/22

hdgarrood:
Yeah, I am a little worried about that. I would quite like to start using some tool which can generate TOCs for us, then I don't think we would need to worry. Splitting syntax up into a few separate pages is also an option but I'm not sure if there's a good way of doing that.

Speaking of tools which generate TOCs, does anyone remember why we stopped using ReadTheDocs?

chexxor:
Regarding doc tools, we can discuss in a thread of its own: https://github.com/purescript/documentation/issues/53

I'd prefer to consider do a structural change, as I suggested in this thread, in a separate PR.

hdgarrood:
Hold on, weren't there some unresolved questions here?

chexxor:
I believe the only relevant one was whether to put the simplest case of matching on one expression. I see that as being the same syntax and concept as matching multiple cases, so it doesn't need a different explanation.

I felt bad that this info isn't available at all for people looking for it in master branch. It is pretty useful as-is, so I decided it should be merged, then improved later when improvement are found.

hdgarrood:
I understand where you're coming from but I'd prefer that you waited for someone else to agree first. Just asking "can we please merge this for now and come back to it later" is fine.

chexxor:
Good idea - I'm new to this. Thanks!

PR #49, opened 2017-03-05
Haskell Differences: Operator Sections

stephencelis:
A small gotcha for me as I've started experimenting.

I do wish deprecated features had compiler errors :) Took me a bit of sleuthing to track this down!

### DIFF
--- a/language/Differences-from-Haskell.md
+++ b/language/Differences-from-Haskell.md
@@ -242,6 +242,22 @@ apply f x = f x
 infixr 0 apply as $
 ```
 
+## Operator Sections
+
+In Haskell, there is syntactic sugar to partially apply infix operators.
+
+```haskell
+(2 ^) -- desugars to `(^) 2`, or `\x -> x ^ 2`
+(^ 2) -- desugars to `flip (^) 2`, or `\x -> 2 ^ 
+```
+
+In PureScript, you can use partial application notation, instead. (Operator sections were removed in version 0.9.)
+
+```purescript
+(2 ^ _)
+(_ ^ 2)
+```
+
 ## Extensions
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
Definitely worth mentioning this here :+1: although I think we generally say that PureScript does have operator sections, albeit with syntax slightly different to Haskell's, so I think it would be good to make some minor wording changes to reflect this.

stephencelis:
Happy to reword anything!

hdgarrood:
That reminds me, we should fully document all the `_` forms, like `case _ of [...]` and `if _ then x else y`; have we done that already?

hdgarrood:
(Not that that needs to happen now / in this PR)

chexxor:
@hdgarrood Created an issue for that: https://github.com/purescript/documentation/issues/52

hdgarrood:
So I think it would be better to say that PureScript does have operator sections, but the syntax is slightly different to Haskell's. That's more in line with what we've written elsewhere, at least. Other than that I think this looks good :+1:

stephencelis:
@hdgarrood Updated with something that hopefully works 😄

hdgarrood:
I'd prefer to avoid the phrase "partial application notation" because I think it suggests that it's a more general thing that works in a wide variety of cases; really this is just one example out of a very select few syntactic forms where we allow underscores for anonymous function arguments like this, and many of those forms don't even involve any partial application. I think just "operator sections" is a better description of them.

stephencelis:
@hdgarrood Ah, I took that description from elsewhere—an old commit message or something. I can remove!

stephencelis:
Updated. I do think it's a bit confusing to ignore `_`'s slightly wider use, though I imagine https://github.com/purescript/documentation/issues/52 will take care of that.

hdgarrood:
Thank you! Yes, it would probably be good to link to these expanded docs from here once they exist.

chexxor:
@stephencelis Thanks for the addition!

PR #54, opened 2017-03-13
Added a section about import hiding

clayrat:


clayrat:
Maybe we should also mention that name in the beginning of `Importing Modules` section then?

hdgarrood:
Yes, I think that would be good.

clayrat:
Any suggestion to reword then? `This` was already there before I edited tho.

hdgarrood:
Oh oops, so it was. In that case I'll merge this and address that later. Thank you!

PR #63, opened 2017-04-11
Use mod instead of % for Int

### LINKED ISSUE #62: Pattern matching - guards not compiling
The [code for guards example](https://github.com/purescript/documentation/blob/master/language/Pattern-Matching.md#guards) doesn't compile:

```purescript
evens :: List Int -> Int
evens Nil = 0
evens (Cons x xs) | x % 2 == 0 = 1 + evens xs
evens (Cons _ xs) = evens xs
```

It gives:

```
Error 1 of 1

  Could not match type

    Int

  with type

    Number


while checking that type t0
  is at least as general as type Number
while checking that expression x
  has type Number
while checking the type of a guard clause
in binding group evens

where t0 is an unknown type
```

pietro909:
closes #62

### DIFF
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -139,7 +139,7 @@ Guards are used to impose additional constraints inside a pattern using boolean-
 ```purescript
 evens :: List Int -> Int
 evens Nil = 0
-evens (Cons x xs) | x % 2 == 0 = 1 + evens xs
+evens (Cons x xs) | x `mod` 2 == 0 = 1 + evens xs
 evens (Cons _ xs) = evens xs
 ```
 


pietro909:
sorry for the typos.

hdgarrood:
Looks good, but I think I'd prefer to use infix notation with backticks here, as that's a little more idiomatic.

pietro909:
Yeah I agree. I'm studying [a book](https://leanpub.com/purescript/read#leanpub-auto-functions-and-records) right know and I just read 4.6:

> This syntax is called infix function application, and any function can be made infix in this way. It is usually most appropriate for functions with two arguments.

hdgarrood:
Great, thanks very much!

PR #64, opened 2017-04-11
Explained TypeSynonymInstance error

stefanholzmueller:
Motivated by http://stackoverflow.com/questions/43344447/type-class-instances-for-type-synonyms-are-disallowed
Feedback is very much welcome!

### DIFF
--- a/errors/TypeSynonymInstance.md
+++ b/errors/TypeSynonymInstance.md
@@ -20,7 +20,7 @@ Since every value that structually matches the type synonym could be considered 
 
 ## Fix
 
-A [newtype](../language/Types.md#newtypes) declaration generates a brand new type instead of introducing a synonym for an existing type. However, the runtime representation is the same. Newtypes can have type class instances declared for them:
+A [newtype](../language/Types.md#newtypes) declaration generates a brand new type instead of introducing a synonym for an existing type. Newtypes can have type class instances declared for them:
 
 ```purescript
 newtype Mass = Mass Number
--- a/errors/TypeSynonymInstance.md
+++ b/errors/TypeSynonymInstance.md
@@ -20,7 +20,7 @@ Since every value that structually matches the type synonym could be considered 
 
 ## Fix
 
-If a type is wrapped in a [Newtype](https://github.com/purescript/documentation/blob/master/language/Types.md#newtypes) instead, it is considered a different type than the underlying type by the type checker. Newtypes can have type class instances declared for them:
+A [newtype](https://github.com/purescript/documentation/blob/master/language/Types.md#newtypes) declaration generates a brand new type; instead of introducing a synonym for an existing type. However, the runtime representation is the same. Newtypes can have type class instances declared for them:
 
 ```purescript
 newtype Mass = Mass Number
--- a/errors/TypeSynonymInstance.md
+++ b/errors/TypeSynonymInstance.md
@@ -14,17 +14,13 @@ instance showMass :: Show Mass where
 
 ## Cause
 
-*Type synonyms* are merely aliases (usually created to avoid verbose type annotations) and do not declare a new unique type. It is still possible to use the synonym and its equivalent type interchangeably:
+[Type synonyms](https://github.com/purescript/documentation/blob/master/language/Types.md#type-synonyms) are merely aliases (usually created to avoid verbose type annotations) and do not declare a new unique type. It is still possible to use the synonym and its equivalent type interchangeably.
 
-```purescript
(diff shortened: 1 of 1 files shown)


chexxor:
Well-written! Thanks so much for contributing!

errors/TypeSynonymInstance.md
```diff
@@ -3,19 +3,43 @@
...
+  show mass = "Mass: " <> show mass <> "kg"
 ```
 
 ## Cause
 
-Explain why a user might see this error.
+*Type synonyms* are merely aliases (usually created to avoid verbose type annotations) and do not declare a new unique type. It is still possible to use the synonym and its equivalent type interchangeably:
+
+```purescript
+velocity = 123.456 :: Mass  -- oops
```
chexxor: I would change `-- oops` to `-- velocity *is* a number, but not a mass - oops`.

errors/TypeSynonymInstance.md
```diff
@@ -3,19 +3,43 @@
...
+```
+
+Since every value that structually matches the type synonym could be considered of the same type, it is not allowed to implement type classes based on type synonyms.
 
 ## Fix
 
-- Suggest possible solutions.
+If a type is wrapped in a **newtype** instead, it becomes a different type than the underlying type. Newtypes can be used as type class instances:
```
LiamGoodacre: I'd argue that *"it becomes a different type"* and *"Newtypes can be used as type class instances"* don't read very well.

Rewording?
> A **newtype** declaration generates a brand new type; instead of introducing a synonym for an existing type.  However, the runtime representation is the same.  Newtypes can also have type class instances declared for them - they will not inherit the type class instances of the underlying type, but these can be selectively derived via `deriving newtype instance`.

stefanholzmueller: I agree, *becomes* is a poor choice of words. I like "Newtypes can also have type class instances declared for them", thanks for the suggestion.

errors/TypeSynonymInstance.md
stefanholzmueller: Actually, I thought of velocity not as a number, but as another physical quanity. Any idea how to express that more clearly?

stefanholzmueller:
BTW, not sure if there is a "style guide" for documenting errors.
Also, maybe I should link related concepts (type synonyms, type classes, newtypes) somewhere?
Going to sleep now, thanks for the feedback so far.

errors/TypeSynonymInstance.md
chexxor: Velocity is a complex metric, like m/s. Might want to use a simple metric, like meters.

Something like the following, though I think you can take parts of it, as it's relatively long, and maybe not necessary to fully explain.

```
distanceToHome = 123.456 :: Mass
-- Distance is meters, which is one way to use a Number,
--   but it's certainly not a mass. This could easily become a bug.
-- Can fix making it a distinct type, instead of an alias for Number.
```

paf31: Maybe just get rid of the example? Is it really relevant to the error?

chexxor:
Just the template, AFAIK. A doc style guide is a thing we should add at some point, yeah, but don't have right now.

errors/TypeSynonymInstance.md
hdgarrood: @liamgoodacre's suggested rewording looks good to me, except the semicolon should be removed.

hdgarrood:
Oops, didn't meant to submit that half complete sentence.

errors/TypeSynonymInstance.md
stefanholzmueller: It is probably unclear and distracting. I'm leaving it out.

errors/TypeSynonymInstance.md
stefanholzmueller: Used @LiamGoodacre's rewording except for the last sentence. It would be good as an extra example, including code. But I guess if someone gets this error, it is because they tried to declare an explicit type class instance, and are not using `deriving`. Or maybe it belongs into the "Notes" section...?

stefanholzmueller:
I linked directly to the language documentation now, instead of formatting relevant concepts with italic and bold.
Any more suggestions?

hdgarrood:
Looks great, thanks! I'd just suggest using relative links as this might want to move someday.

stefanholzmueller:
No more ideas what to improve. I intentionally left out the note about newtype deriving.

chexxor:
Thanks for your contribution!

PR #69, opened 2017-04-23
Small fixes for `let` and `do` in Syntax.md

andreasthoelke:
Small fixes in the let and do documentation.

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -402,7 +402,7 @@ conditional = if 2 > 1 then "ok" else "oops"
 
 ## Let and where bindings
 
-The `let` keyword defines a collection of local declarations, which may be mutually recursive, and which may include type declarations:
+The `let` keyword introduces a collection of local declarations, which may be mutually recursive, and which may include type declarations:
 
 ``` purescript
 factorial :: Int -> Int
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -402,7 +402,7 @@ conditional = if 2 > 1 then "ok" else "oops"
 
 ## Let and where bindings
 
-The `let` keyword creates an expression with local bindings, which may be mutually recursive, and which may include type declarations:
+The let keyword introduces a collection of local declarations, which may be mutually recursive, and which may include type declarations:
 
 ``` purescript
 factorial :: Int -> Int
@@ -415,7 +415,7 @@ factorial =
     go 1
 ```
 
-The `where` keyword can be used at the end of a value declaration to set up bindings that are visible throughout that value:
+The where keyword can also be used to introduce local declarations at the end of a value declaration:
 
 ``` purescript
 factorial :: Int -> Int
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -402,7 +402,7 @@ conditional = if 2 > 1 then "ok" else "oops"
 
 ## Let and where bindings
 
-`let` expressions introduce a list of local declarations, which may be mutually recursive, and which may include type declarations:
+`let` expressions encompass a list of local declarations, which may be mutually recursive, and which may include type declarations:
 
 ``` purescript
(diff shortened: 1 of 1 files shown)


language/Syntax.md
```diff
@@ -402,7 +402,7 @@ conditional = if 2 > 1 then "ok" else "oops"
 
 ## Let and where bindings
 
-The `let` keyword a collection of local declarations, which may be mutually recursive, and which may include type declarations:
+The `let` keyword defines a collection of local declarations, which may be mutually recursive, and which may include type declarations:
```
hdgarrood: I think I'd prefer "introduces" rather than "defines" here; what does everyone else think?

chexxor:
I would only change it if defines is wrong. Thanks for fixing this, BTW!

hdgarrood:
I think "defines" is wrong &mdash; the keyword can't really define anything by itself, it's just a keyword. I guess I was really asking if anyone can think of any better alternatives to my suggestion.

andreasthoelke:
I agree that "defines" is not correct.  
"The `let` keyword introduces a collection of local declarations, .." is fine, I think.

hdgarrood:
@andreasthoelke Thanks very much! This looks good to merge now as far as I can tell, was there anything else you think wants doing or should I go ahead and merge?

andreasthoelke:
When reading the descriptions of `let` and `where` in conjunction, I noticed that the current version is not making the difference too clear.

We could go:


`let` expressions introduce a list of local declarations (variable bindings?), which may be mutually recursive, and which may include type declarations:
..

`where` clauses can be added at the end of a function declaration, and allow to bind variables (set up declarations) that are visible within the surrounding function context.

language/Syntax.md
```diff
@@ -402,7 +402,7 @@ conditional = if 2 > 1 then "ok" else "oops"
 
 ## Let and where bindings
 
-The `let` keyword a collection of local declarations, which may be mutually recursive, and which may include type declarations:
+`let` expressions encompass a list of local declarations, which may be mutually recursive, and which may include type declarations:
```
hdgarrood: I prefer the previous version with "The `let` keyword introduces a collection of local declarations...", actually.

language/Syntax.md
```diff
@@ -415,13 +415,13 @@ factorial =
     go 1
 ```
 
-The `where` keyword can also be used to introduce local declarations at the end of a value declaration:
+`where` clauses can be added at the end of a function declaration, and allow to bind variables that are visible throughout the function:
```
hdgarrood: I think this sentence was okay before; what was your reasoning behind this change?

hdgarrood: Specifically, it's incorrect to say "function declarations", you can also do e.g.

```purescript
pi :: Number
pi = 4 * (sum (take 100 (taylorSeries arctan 1.0))
  where
  taylorSeries = [...]
  arctan = [...]
```

It would be incorrect to call this a "function declaration" because it isn't declaring a function; you can use a where clause in any kind of value declaration which is why "value declaration" is preferable. Also, "... allow to bind variables ..." is grammatically incorrect. Most importantly though, I think it would be best to make the structure of this sentence resemble that of the earlier sentence that tells you what the `let` keyword does as much as we reasonably can, so that it's as clear as possible that `let` and `where` do basically the same thing.

hdgarrood:
I'm not really sure what this latest iteration adds over what we had previously, ideally I think we would just add "introduces" to the `let` sentence and leave the `where` sentence as-is. Can we please just do that for now, and then potentially discuss further wording changes in a separate PR? Just because the rest of the changes are clearly ready to go now and everyone agrees on them.

hdgarrood:
Great, thanks very much :)

andreasthoelke:
Ok, I changed this back.

PR #73, opened 2017-04-30
Move lua-purescript to stale

Xandaros:
It is unlikely that I will pull in the latest changes any time soon, so I'm officially calling it stale.
If I ever get the urge to deal with it again, I will probably redo it from scratch with CoreFN or, if it'll be available, the imperative core.


hdgarrood:
Thanks!

PR #74, opened 2017-05-06
Modernize some documentation with syntax additions and details

NasalMusician:
Is it better to be explicit with ``# Control.Monad.Eff.Effect`` or let the reader figure out that ``# Effect`` can be used after importing ``Control.Monad.Eff``?

### DIFF
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -33,6 +33,7 @@ The following pattern types are supported:
 - Record patterns
 - Named patterns
 - Guards
+- Pattern guards
 
 The exhaustivity checker will introduce a `Partial` constraint for any pattern which is not exhaustive.
 By default, patterns must be exhaustive, since this `Partial` constraint will not be satisfied. The error can be silenced, however, by adding a local `Partial` constraint to your function.
@@ -125,8 +126,8 @@ Named Patterns
 Named patterns bring additional names into scope when using nested patterns. Any pattern can be named by using the ``@`` symbol:
 
 ```purescript
-f a@[_, _] = true
-f _ = false
+f a@[_, _] = a
+f _ = []
 ```
 
 Here, in the first pattern, any array with exactly two elements will be matched and bound to the variable `a`.
@@ -134,7 +135,7 @@ Here, in the first pattern, any array with exactly two elements will be matched 
 Guards
 ------
 
-Guards are used to impose additional constraints inside a pattern using boolean-valued expressions, and are introduced with a pipe after the pattern::
+Guards are used to impose additional constraints inside a pattern using boolean-valued expressions, and are introduced with a pipe after the pattern:
 
 ```purescript
 evens :: List Int -> Int
@@ -149,3 +150,43 @@ When using patterns to define a function at the top level, guards appear after a
 greater x y | x > y = true
 greater _ _ = false
 ```
+
+To be considered exhaustive, guards must clearly include a case that is always true. Even though the following makes perfect sense, the compiler cannot determine that is is exhaustive:
+
+```purescript
+compare :: Int -> Int -> Ordering
+compare x y
+    | x > y  = GT
+    | x == y = EQ
+    | x < y  = LT
+```
+
+Either of these will work, since they clearly include a final case:
+
+```purescript
+compare x y
+    | x > y = GT
+    | x < y = LT
+    | otherwise = EQ
+
+compare x y | x > y = GT
+compare x y | x < y = LT
+compare _ _ = EQ
+```
+
+(The name `otherwise` is a commonly-used synonym for `true` used in guards.)
+
+Pattern Guards
+--------------
+
+Pattern guards extend guards with pattern matching, notated with a left arrow. A pattern guard will succeed only if the computation on the right side of the arrow matches the pattern on the left.
+
+For example, we can apply a function `fn` to an argument `x`, succeeding only if
+`fn` returns `Just y` for some `y`, binding `y` at the same time:
+
+```purescript
+bar x | Just y <- fn x = ... -- x and y are both in scope here
+```
+
+Pattern guards can be very useful for expressing certain types of control flow when
+using algebraic data types.
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -23,17 +23,17 @@ Patterns can also be used when introducing functions. For example:
 example x y z = x * y + z
 ```
 
-The following pattern types are supported:
+The following forms can be used for matching:
 
-- Wildcard pattern
+- Wildcard patterns
 - Literal patterns
-- Variable pattern
+- Variable patterns
 - Array patterns
 - Constructor patterns
 - Record patterns
 - Named patterns
-- Guards
-- Pattern guards
+
+Guards and pattern guards are also supported.
 
 The exhaustivity checker will introduce a `Partial` constraint for any pattern which is not exhaustive.
 By default, patterns must be exhaustive, since this `Partial` constraint will not be satisfied. The error can be silenced, however, by adding a local `Partial` constraint to your function.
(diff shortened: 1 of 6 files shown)


language/Types.md
```diff
@@ -74,7 +74,16 @@ instance showPercentage :: Show Percentage where
...
+Functions in PureScript are like their Javascript counterparts, but always have exactly one argument. Function declarations and lambdas (anonymous functions) with multiple arguments are shorthand for curried lambda abstractions. For example, the following are all equivalent:
+
+```purescript
+add x y = x + y
+add = \x y -> x + y
+add = \x -> \y -> x + y
+add = (+)
+```
+
+Despite this overwhelming convention, uncurried JavaScript functions are can be used through the [purescript-functions](https://pursuit.purescript.org/packages/purescript-functions/) library. For more information see [The Foreign Function Interface](http://www.purescript.org/learn/ffi/).
```
paf31: "overwhelming"? 😄

NasalMusician: Ugh diction. Never works the first time (unlike PureScript!)

Is this better? (also changed the link to point to the relevant section)
> Since uncurried JavaScript functions can be more efficient to run, they are provided indirectly through the [purescript-functions](https://pursuit.purescript.org/packages/purescript-functions/) library. This is unnecessary, however, in most cases and should **not** be used for APIs. For more information see [Functions of Multiple Arguments](http://www.purescript.org/learn/ffi/#functions-of-multiple-arguments).

language/Types.md
```diff
@@ -152,54 +161,88 @@ since the skolemized type variable ``a`` does not unify with ``Int``.
...
 
 ```purescript
-(name :: String, age :: Number | r)
+( name :: String, age :: Number | r )
 ```
 
 ## Type Synonyms
 
-For convenience, it is possible to declare a synonym for a type using the ``type`` keyword. Type synonyms can include type arguments.
+For convenience, it is possible to declare a synonym for a type using the ``type`` keyword. Type synonyms can include type arguments but [cannot be partially applied](https://github.com/purescript/documentation/blob/master/errors/PartiallyAppliedSynonym.md#partiallyappliedsynonym-error). Type synonyms can be built with any other types but [cannot refer to each other in a cycle](https://github.com/purescript/documentation/blob/master/errors/CycleInTypeSynonym.md#cycleintypesynonym-error).
```
hdgarrood: Could you use relative links here please?

language/Types.md
```diff
@@ -152,54 +161,88 @@ since the skolemized type variable ``a`` does not unify with ``Int``.
 
 ## Rows
 
-A row of types represents an unordered collection of named types without duplicates.
+A row of types represents an unordered collection of named types. Duplicate labels are allowed but are distinct.
```
hdgarrood: I think this is maybe a little unclear; possibly we should use the same text as from the release notes here.

language/Types.md
hdgarrood: I think this new paragraph is a bit outside the scope of this file; I'm not sure this should be added here.

language/Types.md
NasalMusician: This?
> Rows are now unordered collections (of labels and types) with duplicates. However, the collection of types for a specific label within a row is ordered. Conceptually, a row can be thought of as a type-level Map Label (NonEmptyList Type).

What about ...
> A row of types represents an unordered collection of named types, with duplicates. Duplicate labels have their types collected together in order, as if in a ``NonEmptyList``. This means that, conceptually, a row can be thought of as a type-level ``Map Label (NonEmptyList Type)``.

Should we introduce the idea that labels can be strings too? (Just noticing that string labels are non-obvious but used in Halogen/DOM.)

language/Types.md
NasalMusician: Well … I think the statement needs *some* explanation for those who do not have functional programming backgrounds (the heresy! 😉 ). So we at least need to explain like, No guys, you really can give multiple arguments to functions!! It just works a bit differently. And at the type level we pretend like everything takes one argument. Maybe a link to a good explanation of currying would be appropriate?

I added the link to the library to encourage readers' natural curiosity. I agree, this file's job is not to explain everything, but I do think the different bits of documentation need to be better linked together to encourage exploration beyond "functions take a single argument". Because what are these types but functions that take more than one argument? Plus, even though they're supported by the compiler, it's all basically just JS/PS code running it! I actually think this library is a prime example of PureScript's flexibility and it would be awesome to encourage learners to be curious about all of this.

Thanks for the feedback, I will keep pondering it.

language/Pattern-Matching.md
```diff
@@ -33,6 +33,7 @@ The following pattern types are supported:
 - Record patterns
 - Named patterns
 - Guards
+- Pattern guards
```
hdgarrood: Pattern guards are not a type of pattern, they're a context in which patterns can be used, so I don't think they should be in this list.

language/Types.md
```diff
@@ -152,54 +176,88 @@ since the skolemized type variable ``a`` does not unify with ``Int``.
...
+
+-- Create an alias for a complex function type
+type Baz = Number -> Number -> Bar Number
+
+-- This function will take two arguments and return a record with double the value
+mkDoubledFoo :: Baz
+mkDoubledFoo foo bar = { foo: 2.0*foo, bar: 2.0*bar }
+
+-- Build on our preious functions to double the values inside any Foo
+-- (Rembmer that Bar Number is the same as Foo)
```
hdgarrood: Typo here?

language/Pattern-Matching.md
hdgarrood: I suppose "Guards" shouldn't really be there either, actually, in that case..?

NasalMusician: Is there any place where patterns are allowed but guards are not? View patterns in Haskell even desugar to guards, so I would lean towards saying they are _at least_ intimately related concepts.

hdgarrood: Yes, of course I agree that they are intimately related, and it's worth having discussion of them in this file; I just think it's incorrect to include them in this list.

NasalMusician: Ah, sorry I misunderstood. I personally cannot think of a better way to integrate them, though. Would you like a note afterward like "Guards and pattern guards are also supported"?

I could also hyperlink them, that might be nice …

language/Types.md
NasalMusician: Thanks

language/Types.md
NasalMusician: Okay ... What about a very short note with two links? Or just revert the changes?

> Functions in PureScript (defined as `input -> output`) are like their JavaScript counterparts, but are curried to always have exactly one argument. Currying works just like [Haskell](https://wiki.haskell.org/Currying).
>
> Uncurried JavaScript functions are also available through the [purescript-functions](https://pursuit.purescript.org/packages/purescript-functions/) library.

The reason I include the link to Haskell is that it explains several features of currying quite well: function returns function, partial application, associativity, pointfree, etc.

hdgarrood: Yes, of course we do need to explain this somewhere; I just think it's important to make sure this particular part of our documentation is focused, to keep it concise and to make sure that people don't have to wade through stuff they already know to get to what they're interested in. If they're here, they probably know what they're looking for (remember that this part of the repo is a language reference, and that is how language references are meant to be used). We have PureScript by Example for people who don't have FP backgrounds or indeed anyone new to PureScript - currying and the purescript-functions library are both covered there.

I do think it would be best to revert these changes - although I'd like to hear from other maintainers too.

language/Pattern-Matching.md
hdgarrood: That sounds good to me, yeah. @paf31 / @garyb, thoughts?

NasalMusician:
Sorry for the mess, should I squash it and create a new PR? Or I could categorize/separate out the changes better.

PR #75, opened 2017-05-09
Update kinds, add newer syntaxes, notes

NasalMusician:
Replaces #74, just cleaned up the ugly commits.

### DIFF
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -174,7 +174,7 @@ compare x y | x < y = LT
 compare _ _ = EQ
 ```
 
-(The name `otherwise` is a commonly-used synonym for `true` used in guards.)
+(The name `otherwise` is a synonym for `true` commonly used in guards.)
 
 Pattern Guards
 --------------
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -135,7 +135,7 @@ Here, in the first pattern, any array with exactly two elements will be matched 
 Guards
 ------
 
-Guards are used to impose additional constraints inside a pattern using boolean-valued expressions, and are introduced with a pipe after the pattern::
+Guards are used to impose additional constraints inside a pattern using boolean-valued expressions, and are introduced with a pipe after the pattern:
 
 ```purescript
 evens :: List Int -> Int
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -126,8 +126,8 @@ Named Patterns
 Named patterns bring additional names into scope when using nested patterns. Any pattern can be named by using the ``@`` symbol:
 
 ```purescript
-f a@[_, _] = true
-f _ = false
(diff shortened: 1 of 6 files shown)


language/Types.md
```diff
@@ -152,54 +152,88 @@ since the skolemized type variable ``a`` does not unify with ``Int``.
...
-x = one :: forall a. (Semiring a) => a
-x = one :: Int
+-- Defined in Data.Semiring
+one :: forall a. (Semiring a) => a
+
+-- This can be constrained to just an Int, since Int is an instance of Semiring
+int1 :: Int
+int1 = one -- x = 1
+-- Or even a Number, which also provides a Semiring
+number1 = one :: Number -- y = 1.0
```
LiamGoodacre: What are `x` and `y` here?

language/Pattern-Matching.md
```diff
@@ -23,16 +23,17 @@ Patterns can also be used when introducing functions. For example:
 example x y z = x * y + z
 ```
 
-The following pattern types are supported:
+The following forms can be used for matching:
```
hdgarrood: Can you explain the rationale behind this change?

chexxor: I like this change, as "form" has fewer connotations than "type", which could confuse newcomers.

language/Types.md
NasalMusician: Typo, thanks for noticing! Should be `int1`/`number1` as I wanted to provide the concrete value defined for each `one`.

chexxor: It seems to be way to explain to a newcomer what `one` is, here. I think this can be expressed more clearly, perhaps:

`int1 = one -- evaluates to 1`
`number1 = one :: Number -- evaluates to 1.0`

A thought on this example: This example requires the reader to understand typeclasses and that `Int` and `Number` are `Semiring`. This might be a unnecessarily large prerequisite as an example of type annotations, but we can find a simpler example later. A imagine a simpler example would be a simple type variable:

```
x = 1 -- x :: Int
x = 1.0 -- x :: Number
x = (1+_) -- x :: Int -> Int
x = (+) -- x :: Semiring a => a -> a -> a
```

language/Pattern-Matching.md
NasalMusician: Yes, what he said :). I decided on that wording to draw the distinction we discussed between guards, which are conditional in nature, and what I would call forms of patterns, which match against an argument's shape or value.

hdgarrood: Ok, :+1: for this then

NasalMusician: Okay, thanks for your attention to detail!

chexxor:
LGTM

hdgarrood:
Thanks very much!

NasalMusician:
Thanks for the feedback!

paf31:
Thanks!

PR #76, opened 2017-05-13
Some re-organization of README.md and Tools.md

soupi:
[View it here](https://github.com/soupi/documentation).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,51 +1,43 @@
-Welcome to the PureScript wiki! This wiki is meant to be a collaborative effort, so please feel free to add/edit content where it is appropriate to do so.
+![PureScript](https://github.com/purescript/purescript/raw/master/logo.png)
+
+Welcome to the PureScript documentation repository! This repository is meant to be a collaborative effort, so please feel free to add/edit content where it is appropriate to do so.
 
 ## Introduction
 
-PureScript is a small strongly, statically typed compile-to-JS language with a number of interesting features, such as:
+PureScript is a small strongly typed programming language that compiles to JavaScript.
+To get a better overview of PureScript, visit [The PureScript Website](http://purescript.org).
 
-- Type Inference
-- Higher Kinded Polymorphism
-- Support for basic JavaScript types
-- Extensible records
-- Extensible effects
-- Optimizer rules for generation of efficient JavaScript
-- Pattern matching
-- Simple FFI
-- Modules
-- Rank N Types
-- Do Notation
-- Tail-call elimination
-- Type Classes
-- Functional Dependencies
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,51 +1,42 @@
...
+PureScript is a small strongly typed programming language that compiles to JavaScript.
+To get a better overview of PureScript, visit [The PureScript Website](http://purescript.org).
 
-- [Maintained Packages](ecosystem/Maintained-Packages.md)
-- [Style Guide](guides/Style-Guide.md)
-- [Related Projects](language/Related-Projects.md)
+## Getting Started
+
+- [Getting Started guide](http://www.purescript.org/learn/getting-started/): Download PureScript and build your first project
+- [Try PureScript](http://try.purescript.org): Try PureScript in your browser
```
hdgarrood: Perhaps PureScript by Example should go here too?

README.md
```diff
@@ -62,10 +53,10 @@ The [PureScript Book](https://leanpub.com/purescript/read) is the recommended ap
...
 - [Differences from Haskell](language/Differences-from-Haskell.md)
-- [Alternate compiler backends](ecosystem/Alternate-backends.md) for various target languages
 
 ## TL;DR
-* [Install PureScript](http://www.purescript.org/download/)
-* [Install Node.js](https://nodejs.org/): `brew install node`/`pacman -S nodejs`/`choco install nodejs`/[Download Node.js](https://nodejs.org/)
-* [Install Pulp](https://github.com/bodil/pulp): `npm install -g pulp bower`
-* GO: `pulp init`, `bower install --save purescript-tuples`, `pulp build --to output.js`, etc.
+
+- [Install PureScript](http://www.purescript.org/download/)
```
hdgarrood: Maybe let's just remove everything under the TL;DR heading? The getting started guide didn't exist when this was written; now that it does I'm not sure this stuff should stay here. If people don't know this stuff already we should direct them to the getting started guide, because it's better maintained and easier to find.

soupi:
Made the changes as requested.

> The only thing is I can't comment on the build tools stuff for editors other than Vim, so it would be good to get people who use each of those editors to look over those bits.

Yeah, I'd appreciate it as well.

paf31:
👍 I like it!

paf31:
Oh also - I think maybe we should test whether the Nix article is still up-to-date.

soupi:
btw I included the Nix article because it's already in the current README, I have no problem removing it if that's what you prefer but I currently don't have the time to check if it still works or not. sorry.

kRITZCREEK:
Thanks!

PR #77, opened 2017-05-31
Clean up Emacs section

kRITZCREEK:
Removes obsolete `purscheck` reference, makes it clear how the different packages correspond to one another. Add a note about spacemacs


paf31:
👍 Looks good!

PR #79, opened 2017-06-12
Update references to old binary `psc`

hdgarrood:
This is just in the Getting Started guide, I should also grep the entire repo for any more of these which might still be hanging around.


paf31:
👍 Great, thanks!

PR #83, opened 2017-06-22
Add example of unknown module error

tchajed:
Describe possible fix that you need to install a dependency with
`bower`.


hdgarrood:
Thanks! Just one thing: I don't think this is quite the right wording because not everyone uses bower. I think it would probably be best to put something like "make sure your dependencies have been installed" and not refer to any particular package manager.

tchajed:
Sounds good, fixed.

hdgarrood:
Fab, thank you!

Issue #86, opened 2017-06-25
Getting started: let in PSCi


i3games:
The "Getting Started" tutorial uses `let` twice in PSCi. This throws an error, as 
`Declarations in PSCi no longer require "let", as of version 0.11.0`

PR #89, opened 2017-06-25
Remove "let" from psci commands.

chexxor:
Should we fix this same issue in other places in this repo? I did a search and found lots of places.

Fixes https://github.com/purescript/documentation/issues/86


chexxor:
https://github.com/purescript/documentation/pull/90 fixes the issue better than my PR. I'll leave this open, however, to fix this PSCI change in other places.

hdgarrood:
Since the getting started guide has now been fixed I'll close this, but I've opened #98 instead.

Issue #88, opened 2017-06-25
Getting started: more warnings about missing type declarations


i3games:
Invoking [pulp build](https://github.com/purescript/documentation/blob/master/guides/Getting-Started.md#compiling-a-solution) throws 3 more `missing type declarations` warnings #87 for Euler.purs

hdgarrood:
Fixed by https://github.com/purescript/documentation/commit/3b20d93be82b2f6b65645e2108816f879d858539

Issue #87, opened 2017-06-25
Getting started: no type declaration for main


i3games:
[invoking `pulp run`](https://github.com/purescript/documentation/blob/master/guides/Getting-Started.md#creating-executables) in the Getting Started guide throws a warning ` No type declaration was provided for the top-level declaration of main....`

chexxor:
We can fix this by adding the type signature to that function, something like `main :: forall e. Eff (console :: CONSOLE | e) Unit`, but that might require explaining `Eff`, which doesn't belong in "Getting Started".

@i3games How do you propose this be fixed?

i3games:
I just added pull request #90 which does add type declarations. If a compiler complains about something by default, I think it is better to confront beginners with learning about these concepts instead of saying "ignore these warnings".

chexxor:
I can accept that reasoning. We should wait for other people to review this solution before accepting it.

@i3games Thanks so much for reporting these issues, and even fixing them!

i3games:
You're welcome. I'd like to help out with documentation in the future as I'm getting more into PS.

PR #90, opened 2017-06-25
Getting Started tutorial

### LINKED ISSUE #86: Getting started: let in PSCi

i3games:
Getting Started: removing let, adding type declarations (fixes #86 #87 #88) and in the test file.

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -182,10 +182,7 @@ Alternatively, we can use Pulp to compile our new module to JavaScript:
 
 This will compile each module present in `src/` into a separate file in the `output/` directory.
 
-The compiler will display several warnings about missing type declarations. Type declarations
-are an advanced topic, therefore they are left out from this guide. In general it is considered
-good practice to provide explicit type signatures. In their absence, the PureScript compiler
-is able to infer types automatically but will remind us to consider adding them.
+The compiler will display several warnings about missing type declarations. In general it is considered good practice to provide explicit type signatures. In this guide, they are left out for brevity. In the absence of type signatures, the PureScript compiler infers types automatically but will remind us to consider adding them.
 
 #### Writing a Test Suite
 
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -157,13 +157,16 @@ module Euler where
 
 import Prelude
 
-import Data.List (range, filter)
+import Data.List (List, range, filter)
 import Data.Foldable (sum)
 
+ns :: List Int
 ns = range 0 999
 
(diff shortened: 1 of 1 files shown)


guides/Getting-Started.md
```diff
@@ -196,8 +199,10 @@ module Test.Main where
 import Prelude
 
 import Euler (answer)
-import Test.Assert (assert)
+import Control.Monad.Eff (Eff)
+import Test.Assert (assert, ASSERT)
 
+main :: forall e. Eff (assert :: ASSERT | e) Unit
```
paf31: I think this might be a bit confusing as part of a getting started guide. We could instead mention that the compiler will infer the type and we can choose to add it based on the warning. Most editor plugins provide a way to add the inferred type annotation.

i3games: I took them out again and wrote a paragraph explanation instead.

hdgarrood:
I'd actually prefer we just put the type signatures in, perhaps with a comment that the reader shouldn't worry about precisely what they mean yet. In particular I imagine few will struggle with signatures for simpler types, e.g. `ns :: Array Int` or `answer :: Int` a little further up. We could also use a closed row for the `Eff` ones to make them slightly less intimidating. Whatever we do please let's not describe type signatures as an advanced topic.

paf31:
I just worry that `Eff (console :: CONSOLE) Unit` is a complex type for a first-time user and it might put some users off. It's not an advanced topic, but it's certainly more advanced than `main = log "Hello World"`.

If we do include the type signature, I'd like to at least say something like "don't worry if you don't understand this type yet, the compiler will infer it anyway".

paf31:
Ah I read the updated description. Yes, I agree we shouldn't say type annotations are an advanced topic. Maybe we can just reword it slightly.

hdgarrood:
Ok, rewording it slightly sounds fine to me then. :+1:

i3games:
Reworded it slightly.

guides/Getting-Started.md
```diff
@@ -196,7 +198,7 @@ module Test.Main where
 import Prelude
 
 import Euler (answer)
-import Test.Assert (assert)
+import Test.Assert (assert, ASSERT)
```
hdgarrood: Let's revert the change to this line this since it's no longer necessary.

i3games: Done.

hdgarrood:
Looks great, thank you!

PR #92, opened 2017-06-26
Document guards for 'case' expressions

geigerzaehler:
~~Also includes TOC for `Syntax.md`.~~

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -424,6 +424,17 @@ f a b = case a, b of
 f (Just true) (Right true)
 ```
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
TOCs are nice but I'd rather not include them like this: without a solid mechanism for keeping them up to date it most likely won't happen. Also could you add a comma after "Like top-level declarations" please?

geigerzaehler:
Mind having another look at @hdgarrood?

hdgarrood:
Thanks!

PR #95, opened 2017-07-03
Tweaks to getting started guide

hdgarrood:
* Use current Show List instance
* Avoid potentially confusing use of the phrase "let binding" (since the
  user no longer will actually type the word "let" into the repl)
* Avoid mixup of lists and arrays


paf31:
👍 Thanks!

PR #101, opened 2017-07-08
Update repl sessions to avoid obsolete 'let'

### LINKED ISSUE #98: Search for and remove "let" in psci commands
See #89

hdgarrood:
As of v11.0.0, local bindings in a repl session must be written without
a 'let', e.g. 'x = y'. Fixes #98


paf31:
Thanks again!

PR #103, opened 2017-07-08
Add section on Deriving to the Differences from Haskell page

cdepillabout:
This PR adds a section on "deriving" to the Differences from Haskell page.

It was inspired from https://github.com/purescript/purescript/issues/2975#issuecomment-313671828.

### DIFF
--- a/language/Differences-from-Haskell.md
+++ b/language/Differences-from-Haskell.md
@@ -163,7 +163,7 @@ Overlapping instances are still disallowed, like in Haskell. The instance names 
 
 ### Orphan Instances
 
-Unlike Haskell, orphan instances are completely disallowed in Purescript.  It is a compiler error to try to declare orphan instances.
+Unlike Haskell, orphan instances are completely disallowed in PureScript.  It is a compiler error to try to declare orphan instances.
 
 When instances cannot be declared in the same module, one way to work around it is to use [newtype wrappers](http://stackoverflow.com/questions/22080564/whats-the-practical-value-of-all-those-newtype-wrappers-in-data-monoid).
 
--- a/language/Differences-from-Haskell.md
+++ b/language/Differences-from-Haskell.md
@@ -179,21 +179,17 @@ derive instance eqFoo :: Eq Foo
 derive instance ordFoo :: Ord Foo
 ```
 
-The following type classes can be derived:
+Examples of type classes that can be derived this way include `Eq`, `Functor`,
+and `Ord`.  See
+[here](https://github.com/purescript/documentation/blob/master/language/Type-Classes.md#type-class-deriving)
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think it would be better to link to the existing list of derivable classes rather than trying to maintain this list in two separate places.

hdgarrood:
Come to think of it, re the Generics part, a) does it belong there? I'm not sure, and b) if it does, would it work to just link to the generics-rep docs on Pursuit and say 'see here for a list of classes you can define instances for using generics'?

paf31:
The problem is that there isn't a closed set of things you can derive with `generics-rep`, so I think we should make it clear that these are just a few examples, and that many libraries provide other generic-derivable classes.

cdepillabout:
As someone coming from Haskell to PureScript, I really like having a page that answers answers the question "How do I do _x_ in PureScript?"

In Haskell, I very often derive things like `Eq`, `Generic`, `Ord`, `Show`, etc.  Since PureScript doesn't have the `deriving` keyword, I think it is really helpful to show how things can be easily derived in PureScript, whether that's by the compiler or generic deriving.

That being said,

> I think it would be better to link to the existing list of derivable classes rather than trying to maintain this list in two separate places. Come to think of it, re the Generics part, a) does it belong there? I'm not sure, and b) if it does, would it work to just link to the generics-rep docs on Pursuit and say 'see here for a list of classes you can define instances for using generics'?

I agree that it might be better to keep these lists in one place instead of spreading them over multiple pages.

> The problem is that there isn't a closed set of things you can derive with `generics-rep`, so I think we should make it clear that these are just a few examples, and that many libraries provide other generic-derivable classes.

Coming from Haskell, I would assume most people would realize this without being explicit about it, but that doesn't mean it isn't a good idea to be explicit.

cdepillabout:
In commit f9a1a06, I've fixed the explanation based on the feedback from @hdgarrood and @paf31.

I've taken out the lists of type classes and instead just linked to the pages describing the features in question.

However, I have left in some examples of type classes that can either be derived by the compiler or with generics.  This is for someone coming from Haskell, opening up this page, and just doing a Ctrl-F for "Show" or "Eq".

In Haskell, it is so common to just derive things like `Show` and `Eq` that I feel like those strings should definitely appear in this page.

cdepillabout:
@hdgarrood 
> would it work to just link to the generics-rep docs on Pursuit and say ...

In the generics section, I'm currently linking to https://github.com/purescript/documentation/blob/master/guides/Generic.md, but do you think it would be better to link directly to `generics-rep`?  I was conflicted by which to do.

I guess it would be possible to link to both, but I wasn't sure how to word it.

PR #106, opened 2017-07-31
Replace psci with repl in command line

wuct:
Actually, I don't know if there is any difference between `pulp psci` and `pulp repl`, but there is only `pulp repl` listed under `pulp --help` in pulp 11. So to avoid confusion, I believe it better to remove `pulp psci` in the doc.

### DIFF
--- a/errors/CycleInDeclaration.md
+++ b/errors/CycleInDeclaration.md
@@ -3,7 +3,7 @@
 ## Example
 
 ```text
-$ pulp psci
+$ pulp repl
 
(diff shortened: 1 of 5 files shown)


hdgarrood:
Yes, `pulp psci` is just for backwards compatibility really. Thanks!

wuct:
Thanks for merging

Harry Garrood <notifications@github.com>於 2017年7月31日 週一，下午7:16寫道：

PR #111, opened 2017-08-19
included atom plugin under Editor & tools section

nikhilnagaraju:
Made a simple Atom plugin which helps to quickly refer Purescript modules , functions or documentation on Pursuit


nikhilnagaraju:
@paf31 , @hdgarrood please review the PR, If its merged it will be helpful to other developers too. Thank you

hdgarrood:
Thanks for the PR. You should really wait at least a day or two before bumping, though.

Does your plugin do anything that the existing one doesn't? If the answer to that is yes, would it make sense to implement the new stuff inside the existing plugin? I would just rather keep these lists as short as possible.

PR #123, opened 2017-10-05
Fill in the InvalidInstanceHead documentation

i-am-tom:
Previously, there wasn't a lot of information around this error, which was a little scary, given how often it's popping up now that RowToList is de rigeur. Here's my attempt at helping.

Let me know what you think, and whether I should make a point of including worked examples in the error pages!

### DIFF
--- a/errors/InvalidInstanceHead.md
+++ b/errors/InvalidInstanceHead.md
@@ -2,20 +2,129 @@
 
 ## Example
 
-```purescript
-module ShortFailingExample where
+As of 0.11.6, the `InvalidInstanceHead` error has been added to the
+purescript compiler. The error can be summoned using the following:
 
+```purescript
+module InvalidInstanceHead where
+
+class ClassWithInvalidInstanceHead (x :: # Type)
+
+instance FailingExample :: ClassWithInvalidInstanceHead ()
 ...
 ```
 
 ## Cause
 
-Explain why a user might see this error.
+Normally, when this error rears its head, it's because you're trying to write
+an `instance` of a given `class` that pattern matches on some **row type**. The
+issue here is that there isn't really a sensible way to match on records. There
+are, however, ways to get around this limitation.
 
 ## Fix
 
-- Suggest possible solutions.
+With the arrival of [`purescript-typelevel-prelude`](https://github.com/purescript/purescript-typelevel-prelude)
+2.3.0 came the introduction of `RowToList` and its surrounding classes. The
+basic idea is that you can convert a row type to a type-level cons list of its
+entries. For example:
 
-## Notes
+```purescript
+--        ROW              LIST
 
-- Additional notes.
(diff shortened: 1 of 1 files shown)


errors/InvalidInstanceHead.md
```diff
@@ -2,20 +2,129 @@
...
+
+```purescript
+instance isHomogenousImpl
+  :: ( RowToList rowType rowList
+     , IsHomogenousRow rowList a
+     )
+  => IsHomogenousRecord rowType
+```
+
+Here, we say that, "if you can find me a `RowList` representation for my record
```
chexxor: "record" -> "row" (in a few places here, because there's no records here, just rows)
I don't see "la" anywhere, so you might want to update "we don't know what `la` is"

errors/InvalidInstanceHead.md
```diff
@@ -2,20 +2,129 @@
...
+that satisfies `IsHomogenousRow`, then this record is good to go". Notice how
+we're "pattern matching" on the left side of `RowToList` - we don't know what
+`la` is, but we trust the compiler to iron out that little detail for us.
+
+Now that's all ready to roll, we should probably write `IsHomogenousRow`:
+
+```purescript
+class IsHomogenousRow (list :: RowList) a | list -> a
```
chexxor: What do you think about this?
"IsHomogenousRow" -> "IsHomogenousRowList"

Also, would be really nice to say why we need the "a" above. It looks like "a" is the type of the homogenous row - that is, every value in the row needs to be type "a".

Maybe put this above:
"Now we implement `IsHomogenousRow`. The type parameter `list` is the RowList we want to check has homogeneously typed labels, and the type parameter `a` is that one type."

errors/InvalidInstanceHead.md
```diff
@@ -2,20 +2,129 @@
...
+```
+
+Because we're now dealing with a `RowList`, we can pattern match and avoid those
+pesky errors about instance heads. Our functional dependency here says that, "if
```
chexxor: I think this explanation of the functional dependency can be deleted - it isn't any more informative that the definition of a functional dependency, which can be found elsewhere in this documentation or wikipedia.

kRITZCREEK: I don't mind the explanation here. I've definitely found myself chasing definitions until I didn't know where I started anymore before.

kRITZCREEK:
@i-am-tom I think this looks great, I think @chexxor's comments all seem like improvements except for the explanation of functional dependencies, so maybe we could address those and get this merged? Thanks!

i-am-tom:
@kRITZCREEK Yep, sorry about the wait - will get this sorted now :)

i-am-tom:
Apparently deleted my fork, so #163 is the same PR with those issues fixed!

PR #130, opened 2017-10-19
Fix a formatting mistake

sjsyrek:


hdgarrood:
I think this is fine as-is, but thanks anyway!

PR #131, opened 2017-10-19
Fix typos and various other errors

sjsyrek:
I only fixed typos and one grammatical problem. I added clarifying punctuation in a few places. I added missing backticks for consistency (e.g. the Prelude library in line 323). I noted this line:

>* `Monoid` has a superclass `Semigroup`, which provides `(<>)`, and does not require an identity.

Haskell has `Semigroup` now, so you may want to indicate as much, but I didn't go ahead and make this change, myself.


hdgarrood:
I think the semigroup stuff should be left alone for now, maybe later when more people are using a version with semigroup as a superclass we can revisit that.

Issue #124, opened 2017-10-05
No license


paf31:
This is a bit of a disaster from a licensing perspective. As I understand it, the content of every accepted PR is All Rights Reserved by its original author. This means users only have the rights ensured by the GitHub TOS, such as to view, fork, etc. but not to use the content for any purpose, and not to redistribute it.

I would have liked this content to be permissively licensed, but it's a bit late now.

This leaves us with a few options:

- Start over with a fresh license. Obviously this isn't ideal, but I wouldn't be against it if I had more time or more help to rewrite stuff. We might be able to salvage older versions of bits of content if they were moved from old repos with licenses.
- Try to get previous contributors to explicitly license their content like we did with the compiler repo. We know how that went, and it took a long, long time, and was never even finished. Fortunately, I believe the GitHub terms save us there, since we did have a LICENSE file at least, and users are required to license their PR'd code under the same terms. The same does not apply here, however.

Whatever we do, I would like to start requiring explicit licensing in some form for new contributions.

hdgarrood:
I think a combination of these is probably the way to go: try to get previous contributors to explicitly license their content, and rewrite any parts whose authors we can't get in touch with. I would guess that it would only take a few people to explicitly license their contributions to get this repo into a stage where the vast majority of the content is permissively licensed.

Do we maybe want a CONTRIBUTORS.md file like in the compiler?

paf31:
Yes I think that's the way to go.

hdgarrood:
We now have a CONTRIBUTORS.md file which means that most of the content here is now licensed, but in order to fully address this issue we still need someone to go through the contributors tab and/or git history and get consent from any remaining contributors, or alternatively to delete any remaining unlicensed content, to ensure that everything is covered.

PR #134, opened 2017-10-19
Start contributors file

LiamGoodacre:


paf31:
Looks good, thanks, but I think we should mention the contributors which have not added their names here, if only by linking to the contributors tab on GitHub.

CONTRIBUTORS.md
```diff
@@ -0,0 +1,27 @@
...
+
+> My existing contributions and all future contributions until further notice are Copyright {Name}, and are licensed to the owners and users of the PureScript documentation project under the terms of the {License}.
+
+By adding your name to the list below, you agree to license your contributions under these following terms.
+
+If you would prefer to use different terms, please use the section below instead.
+
+| Username | Name | License |
+| :------- | :--- | :------ |
+| [@LiamGoodacre](https://github.com/LiamGoodacre) | Liam Goodacre | [MIT license](http://opensource.org/licenses/MIT) |
```
paf31: Choose whichever license you like, of course, but I'm going to choose a creative commons license, probably the one I use for the book, and I'm going to recommend others do the same, since I think it's more appropriate for written work.

chexxor:
Is this for legal reasons? Or just for "soft" reasons?
I wonder if we *must* have every contributor in this file, for legal reasons, and to make extra efforts to do that.

paf31:
@chexxor The problem is that there never was a LICENSE file in this repo, or on the wiki, or on the website repo. So any content here is essentially All Rights Reserved by its original authors, per the GitHub terms and conditions. Which means there is no way to use it for any other purpose.

We've discussed moving to a creative commons license, but at this point, doing so would require an explicit license of the original content by each author, as I understand it.

We've also been discussing rewriting content instead.

paf31:
See #124

chexxor:
LGTM

paf31:
I think this should indicate very clearly that any content by unlisted authors is _NOT_ explicitly licensed yet.

hdgarrood:
Merged in #150. Thank you!

PR #135, opened 2017-10-22
A number of people have been confused by the KindsDoNotUnify error me…

AidanDelaney:
…ssage wondering which part of the error refers to the expected kind. Addresses purescript bugs #3127, and #2620

I ran into the above confusion.  It may take a while to fix the implementation in PS.  In the meantime, signal the possible confusion to those reading the fine manual.


chexxor:
Seems ok to me. 👍

PR #144, opened 2017-12-06
add documentation for IncorrectAnonymousArgument

csicar:


chexxor:
Do you know if "anonymous argument" is the correct term for this? I know it's called "operator sections" in [language/Syntax.md - Operator sections](https://github.com/purescript/documentation/blob/fc4a9db4b128aa3331e5f990cb1860e59077af31/language/Syntax.md#operator-sections). I'll ask some people in the chatroom.

csicar:
I assumed it was called "anonymous argument" since the error message is named "IncorrectAnonymousArgument".

chexxor:
I wonder if the following syntaxes are also "anonymous argument":
- `\_ -> 1`
- `_.x`
- `{ foo: _, bar: _ }`

I'll make an issue to resolve that question later. 

Regarding whether to use "operator sections" in this error message, it might make sense to phrase it like: "`_` is an anonymous argument inside of the operator section `(_ + 1)`", so in this error message, it would be more clear to say "In an [operator section](https://github.com/purescript/documentation/blob/fc4a9db4b128aa3331e5f990cb1860e59077af31/language/Syntax.md#operator-sections), like `(_ + 1)`, an anonymous argument can be used only once."

chexxor:
I think it's worth mentioning "operator section" in the explanation - it appears to be a special syntax, explicitly called that in the compiler (https://github.com/purescript/purescript/commit/9c29a145b8abb10e1a7615556e81196c542e2337)

chexxor:
Thanks! LGTM! 👍

hdgarrood:
Although the suggested fixes are technically correct, I think if you're faced with this error, and especially if you're not sure what to do, the best approach is usually just to give the problematic argument(s) a name instead. So I think it would be best to list that as a suggested fix first.

hdgarrood:
That is, if the compiler rejects `foo = (_ * _)`, then you should just write it the good old-fashioned way: `foo x y = x * y`.

csicar:
I agree, naming the arguments is definitely the best solution to the problem. I think keeping the original suggestion as a note might help some people understanding, what exactly caused the error.

PR #147, opened 2017-12-14
Add example for `case _ of`

craigdallimore:
I noticed there was not an example for for `case _ of` in the documentation, is this an appropriate way to describe it?

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -434,10 +434,17 @@ f x = case x of
   Right _ -> "Right"
 ```
 
-A binding can be avoided by pattern matching:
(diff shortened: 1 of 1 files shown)


hdgarrood:
We should definitely add an example of this, but I think it would be a bit clearer to say something like "A binding can be avoided by using a single underscore in place of the expression to match on; in this context, the underscore represents an *anonymous argument*."

I think we should also write out the equivalent code, i.e. we should mention that `case _ of {...}` is the same as `\x -> case x of {...}` - just to make it as clear as possible.

craigdallimore:
Thank you, how does this look?

chexxor:
LGTM! Thanks for adding notes of case-syntax's anonymous argument capability!

hdgarrood:
Thank you!

PR #149, opened 2017-12-17
Update Differences-from-Haskell.md

kcsongor:
`DataKinds` (sort of) and `KindSignatures` are supported by PS

### DIFF
--- a/language/Differences-from-Haskell.md
+++ b/language/Differences-from-Haskell.md
@@ -292,7 +292,7 @@ In PureScript, operator sections look a little bit different.
 
 The PureScript compiler does not support GHC-like language extensions. However, there are some "built-in" language features that are equivalent (or at least similar) to a number of GHC extensions. These currently are:
 
-* DataKinds
+* DataKinds (Note that unlike in Haskell, user-defined kinds are open, and they are not promoted, which means that their constructors can only be used in types, and not in values. For more information about the kind system, see https://github.com/purescript/documentation/blob/master/language/Types.md#kind-system)
 * EmptyDataDecls
 * ExplicitForAll
 * FlexibleContexts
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think the way user-defined kinds works in PureScript is different enough from Haskell's `DataKinds` that it's probably best not to list it here..?

kcsongor:
This is a good point, and I thought of this too. As far as I know, the difference is that in PureScript, type constructors of user-defined kinds can't be used as values, but I don't think that's a very important aspect of the extension (though this is just my opinion).

On the other hand, the ability to define kinds (and use literals in types) would be important to communicate somehow.

For me personally, this list was the single most important part of the document, and I know that if I didn't see `DataKinds` on the list, I wouldn't have expected to find this info elsewhere, though again, this is just me. What do you think?

hdgarrood:
What about putting DataKinds on the list, with a short explanation below, plus a link to https://github.com/purescript/documentation/blob/master/language/Types.md#kind-system?

kcsongor:
Yes, that's good, I've updated the PR. Do you think this explanation makes sense? I put it inline, but maybe it would be better below the list...

hdgarrood:
Yes, that looks good, although I think I'd prefer it below the list.

hdgarrood:
What does "open" mean here?

kcsongor:
I meant that you can define a kind, and then add new constructors to it anytime
```hs
foreign import Kind K
foreign import data A :: K
foreign import data B :: K
```

whereas in Haskell, you can't extend an existing kind after it's defined.

hdgarrood:
Ahh, yes, of course, thanks.

hdgarrood:
Thank you!

PR #152, opened 2017-12-21
suggest using generics for CannotDerive error

csicar:
I changed `HeytingAlgebra Bool` to `Show Bool`, becase I think that a more approachable example.

### DIFF
--- a/errors/CannotDerive.md
+++ b/errors/CannotDerive.md
@@ -5,7 +5,7 @@
 ```purescript
 data Bool = True | False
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
Looks good, but I don't want to include the code here, because if that example needs to change we'll almost certainly forget to update it. People can just follow the link.

hdgarrood:
Thanks! I think you're right, I think it is more likely that people will attempt to derive `Show` than `HeytingAlgebra`.

PR #154, opened 2017-12-28
Explain situations in which importing qualified is useful

chexxor:
I have so much trouble entering a codebase which uses `wrap` and `unwrap` because it could so easily have a different meaning inside that package. That is, that package could actually *not* be importing Data.Newtype and actually have defined a completely unrelated function named `wrap` and `unwrap`.

I didn't think of using qualified imports to resolve this. I might make a PR to the purescript-newtype to recommend its users to import the module as qualified.

### DIFF
--- a/language/Modules.md
+++ b/language/Modules.md
@@ -55,7 +55,44 @@ import B (runFoo)
 
 ## Qualified Imports
   
-Modules can also be imported qualified, which means that their names will not be brought directly into scope, but rather, aliased to a different module name. This can be helpful when avoiding naming conflicts:
+Modules can also be imported qualified, which means that their names will not be brought directly into scope, but rather, aliased as a different module name.
+
+Following are some situations in which qualified imports are quite useful.
+
+### Using generically-named functions
+
+``` purescript
+module Main where
+
+import Data.Newtype (wrap, unwrap) as Newtype
+
+newtype WrappedInt = WrappedInt Int
+derive instance newtypeWrappedInt :: Newtype WrappedInt _
+
+test :: WrappedInt
+test = Newtype.wrap 5
+```
+
+The `Data.Newtype` module's `wrap` and `unwrap` functions can be considered relatively ambigous or generic. To clarify *what* is used to wrap the argument, we can import `Data.Newtype` qualified as `Newtype`.
+
+Another example, fictitious this time:
+
+``` purescript
+module Main where
+
+import MyWebFramework as MyWebFramework
+
+main :: Eff (dom :: DOM) Unit
+main = do
+  elem <- queryDomElementById "appContainer"
+  MyWebFramework.run elem
+  -- that is more clear than
+  -- run elem
+```
+
+It isn't clear what the `run` function is doing until we see its module, `MyWebFramework`. To improve readability of this code, we can import and call it qualified - `MyWebFramework.run`.
+
+### Avoiding naming conflicts
 
 ```purescript
 module Main where
@@ -73,6 +110,8 @@ Operators can also be referenced this way:
 ```purescript
 test' = Array.null ([1, 2, 3] Array.\\ [1, 2, 3])
 ```
+
+### Merging modules
 
 Modules can be merged under the same name, but it is best to use explicit imports to avoid conflicts, in case modules would want to import the same name:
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
I'm not hugely keen on this specific example personally, because `Data.Newtype` is one of the modules which I would barely ever import qualified, and when I see it in other people's code it's also usually not imported qualified. I wouldn't be in favour of merging such a PR to `purescript-newtype` either.

I think a better example would be how `Data.Array`, `Data.List`, `Data.Set`, and `Data.Map` (and others) all export a `fromFoldable` function, and in those cases it's better to use e.g. `import Data.Array as Array`.

chexxor:
But `wrap` is just so non-descriptive, it seems like a perfect example of adding a module qualifier greatly improving clarify. I'm not dead-set on having Newtype as *the* example here, but it seems like a really perfect example.

The `fromFoldable` example is a great idea, too, but it's not something I use often, so I wouldn't want to use it as an example. I prefer an example that is widely relatable.

hdgarrood:
You could say the same thing about `bind` and `append`; this is just what happens when you have typeclasses which lots of types can be made instances of. I disagree that `Newtype.wrap` improves clarity tbh.

I recommended `fromFoldable` because it is something I use all the time, more often than `wrap` or `unwrap` actually.

chexxor:
K, the typeclass aspect of `wrap` is a reasonable point. I've updated to use `fromFoldable` as the example.

hdgarrood:
Thanks. To add explanation to my viewpoint here, the core libraries already import `Data.Newtype` unqualified everywhere; it would not be good to have the core libraries and the documentation repo disagree about what's good style.

chexxor:
K, I did a final review of this - you can do the same.

Thanks for the review, @hdgarrood!

garyb:
@chexxor side thing, but regarding `wrap` and `unwrap`: they shouldn't be used IMO, unless you're writing something that is fully generic over `Newtype`. It's easy to accidentally `wrap` or `unwrap` the wrong thing if the types change, so using the explicit constructor and `un WhateverCtor` is the better way to do things normally.

hdgarrood:
Can you give an example of where using `wrap` or `unwrap` might lead to accidents? I understand this argument where you have type class members where a type parameter only appears in e.g. the argument of a function and not the result, or vice versa (e.g. `show`), but I can't really see how that applies here, since `Newtype` has each parameter determine the other.

language/Modules.md
```diff
@@ -55,7 +55,49 @@ import B (runFoo)
...
+a = Map.fromFoldable [ Tuple 1 "a" ]
+
+b :: List Int
+b = List.fromFoldable [ 1 ]
+
+c :: Set Int
+c = Set.fromFoldable [ 1 ]
+```
+
+Several data structure modules have a `fromFoldable` function which can be used to create an instance of that data structure from any other `Foldable` data structure. We can say the `fromFoldable` function name is overloaded by different modules. To clarify which `fromFoldable` function is being used, we can import that module's functions under a qualified name and use it qualified, like `Set.fromFoldable`.
```
hdgarrood: If we're importing all of `Data.Map`, `Data.List`, and `Data.Set` in the example, this really becomes an example of using qualified imports to avoid naming conflicts. I think we should only be using one `fromFoldable` if we want to make the point that it improves readability. Also, I'm not sure we need a whole module for this example; we can just say that if you're reading code and you see `fromFoldable`, you might refer to the import list to see where it comes from.

Also, I don't think the "we can say the `fromFoldable` function name is overloaded by different modules" really adds anything here; the previous sentence is sufficient, I think.

language/Modules.md
```diff
@@ -55,7 +55,49 @@ import B (runFoo)
...
+
+Another example, fictitious and slightly contrived this time:
```
hdgarrood: I wouldn't describe this example as "contrived"

language/Modules.md
```diff
@@ -55,7 +55,49 @@ import B (runFoo)
 
 ## Qualified Imports
   
-Modules can also be imported qualified, which means that their names will not be brought directly into scope, but rather, aliased to a different module name. This can be helpful when avoiding naming conflicts:
```
hdgarrood: I think this first sentence was ok before, actually

chexxor:
Made edits.

Just a confusion I had when trying to read some PS code - just an opinion of mine. I like having suggestions in docs, as it's where I look when I want to know how a language feature was designed to be used and the various problems it is known to help resolve.

Anyways, the question motivated me to improve this doc. I'd like to have this be a thing we can point at to better answer the question if it's posted elsewhere.

hdgarrood:
Thanks!

Issue #158, opened 2018-01-04
Warning pages/skeletons all labelled as error


nwolverson:
All pages under errors/ appear (at a random sample) to be titled 

*`SomeIssue` Error*

I propose those which are warnings rather than errors should be titled

*`SomeIssue` Warning*

chexxor:
Is there a single place in the compiler which lists all warnings? I brief search can't find it.
Is searching for `shouldWarnWith` the best way to find them?

If so, here's a list I made from that search.

- [ ] DuplicateExportRef
- [ ] DuplicateImport
- [ ] DuplicateImportRef
- [ ] DuplicateSelectiveImport
- [ ] HidingImport
- [ ] ImplicitImport
- [ ] ImplicitQualifiedImport
- [ ] OverlappingPattern
- [ ] MissingNewtypeSuperclassInstance
- [ ] MissingTypeDeclaration
- [ ] MissingTypeDeclaration
- [ ] ScopeShadowing
- [ ] ShadowedName
- [ ] ShadowedTypeVar
- [ ] UnnecessaryFFIModule
- [ ] UnusedDctorExplicitImport
- [ ] UnusedDctorImport
- [ ] UnusedExplicitImport
- [ ] UnusedFFIImplementations
- [ ] UnusedImport
- [ ] UnusedTypeVar
- [ ] UserDefinedWarning
- [ ] UnverifiableSuperclassInstance
- [ ] WildcardInferredType

nwolverson:
I don't think there is, as there is no real distinction between error and warning - they are raised in a different way, rather than having errorness/warningness built into the type itself.

The changes in https://github.com/purescript/documentation/pull/159 should check off a bunch of these.

chexxor:
Ok, I checked-off the warnings you added in that MR. Great progress!

PR #159, opened 2018-01-04
Various imports error/warnings entries

nwolverson:


hdgarrood:
This might be a good opportunity to make a start on #158. Also would you mind changing "This error is auto-fixable" to "This warning is auto-fixable"? Finally do you think there's something we can link to which explains what that means?

nwolverson:
Yeah I was considering taking a pass at a bunch of the imports issues. I want to indicate the class of warnings that the compiler provides suggestions to fix which can be applied with purescript-suggest, pscid or editor plugins - I wonder if there's a better wording which can be used consistently. 

Should I maybe create a page in `guides/` which can be linked to that explains these options?

hdgarrood:
I think the wording is fine, I'm just a little concerned about how useful it will be for people who don't have the context.

> Should I maybe create a page in guides/ which can be linked to that explains these options?

That would be great :+1:

nwolverson:
I've added a short error suggestions guide, and linked this, also went through a batch of warnings around imports.

chexxor:
LGTM :+1:

hdgarrood:
Thanks!

PR #160, opened 2018-01-15
Update compiler solvable type classes, Fixes #157

### LINKED ISSUE #157: update "Compiler-Solvable Type Classes"
We are missing classes in this document: https://github.com/purescript/documentation/blob/master/language/Type-Classes.md

chexxor:
https://github.com/purescript/purescript/blob/38f631c71367cffcc973ce0dab2229e81969e161/src/Language/PureScript/TypeChecker/Entailment.hs#L61

There are other classes which are related to these but aren't solved, BTW, like `Equals` and `ListToRow`, but I don't think those belong here.`

### DIFF
--- a/language/Type-Classes.md
+++ b/language/Type-Classes.md
@@ -114,12 +114,22 @@ foo x = x
 
 Currently, the following type classes can be automatically solved:
 
+Symbol-related class
+
 - [`IsSymbol`](https://pursuit.purescript.org/packages/purescript-symbols/3.0.0/docs/Data.Symbol#t:IsSymbol)
+- `ConsSymbol`
 - [`AppendSymbol`](https://pursuit.purescript.org/packages/purescript-typelevel-prelude/2.5.0/docs/Type.Data.Symbol#t:AppendSymbol)
-- [`Equals`](https://pursuit.purescript.org/packages/purescript-typelevel-prelude/2.5.0/docs/Type.Data.Symbol#t:Equals)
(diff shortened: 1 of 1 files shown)


hdgarrood:
I'd prefer to leave `ConsSymbol` out for now, as it's not in a released version yet, and it might end up having a different name (see https://github.com/purescript/purescript/issues/2903). Also `RowCons` doesn't really perform a union (perhaps it would be best not to have descriptions here, actually, as the actual docs are more likely to be kept up to date and accurate).

chexxor:
Yeah, I was curious about ConsSymbol - thanks for the review!

language/Type-Classes.md
```diff
@@ -114,8 +114,21 @@ foo x = x
 
 Currently, the following type classes can be automatically solved:
 
-- `Warn`
-- [`IsSymbol`](https://pursuit.purescript.org/packages/purescript-typelevel-prelude/1.0.0/docs/Type.Data.Symbol#t:IsSymbol)
-- [`CompareSymbol`](https://pursuit.purescript.org/packages/purescript-typelevel-prelude/1.0.0/docs/Type.Data.Symbol#t:CompareSymbol)
-- [`AppendSymbol`](https://pursuit.purescript.org/packages/purescript-typelevel-prelude/1.0.0/docs/Type.Data.Symbol#t:AppendSymbol)
-- `Union`, which computes the union of two rows of types
+Symbol-related class
```
hdgarrood: “classes” here?

hdgarrood:
Thank you!

PR #163, opened 2018-01-30
Fill in the InvalidInstanceHead documentation

i-am-tom:
Previously, there wasn't a lot of information around this error, which was a little scary, given how often it's popping up now that RowToList is de rigeur. Here's my attempt at helping.

This is a recreation of #123 as I apparently deleted the repo some time in the last month >_< It contains fixes for the points that @chexxor rose. Sorry for the wait, @kRITZCREEK 😳

### DIFF
--- a/errors/InvalidInstanceHead.md
+++ b/errors/InvalidInstanceHead.md
@@ -2,20 +2,132 @@
 
 ## Example
 
-```purescript
-module ShortFailingExample where
+As of 0.11.6, the `InvalidInstanceHead` error has been added to the
+purescript compiler. The error can be produced using the following:
 
+```purescript
+module InvalidInstanceHead where
+
+class ClassWithInvalidInstanceHead (x :: # Type)
+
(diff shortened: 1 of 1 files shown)


errors/InvalidInstanceHead.md
```diff
@@ -2,20 +2,132 @@
...
+     )
+  => IsHomogenousRecord rowType
+```
+
+Here, we say that, "if you can find me a `RowList` representation for my record
+that satisfies `IsHomogenousRow`, then this record is good to go". Notice how
+we're "pattern matching" on the left side of `RowToList` - we don't know what
+`la` is, but we trust the compiler to iron out that little detail for us.
+
+Now that's all ready to roll, we should probably write `IsHomogenousRow`:
```
kRITZCREEK: This should also be `IsHomogenousRowList`

i-am-tom: Motherf, thought I'd got them all. Thanks 🤕

Issue #151, opened 2017-12-19
Filling out the CONTRIBUTORS.md file


hdgarrood:
This repository is currently not in the best position with respect to licensing; see #124 for background.

We have now started a CONTRIBUTORS.md file in the same style as the one in the compiler repository, and we are trying to get it filled out a bit, so that as much of this repository as possible is explicitly licensed.

Instead of having you all send PRs, which will inevitably cause lots of merge conflicts, I'd really appreciate it if you could comment on this issue to confirm that you're happy to license your contributions under the terms of [CC BY-NC-SA 3.0](https://creativecommons.org/licenses/by-nc-sa/3.0/deed) (or some other license if you prefer; we recommend CC BY-NC-SA over MIT or any other software licenses you might be more familiar with, because it makes more sense for written work). I (or some other maintainer) can then update the CONTRIBUTORS.md file appropriately.

I've heard GitHub doesn't like it when you tag lots of people at once so I'll just start from the beginning of the Contributors tab: @paf31 @garyb @chexxor @andyarvanitis @jdegoes @epost @joseanpg @jplatte @Fresheyeball @Thimoteus @kRITZCREEK @puffnfresh @alexchandel @nwolverson @dgendill @Xandaros @stefanholzmueller @haf @csicar @damncabbage

jplatte:
CC BY-NC-SA 3.0 sounds good to me! By the way, in GitHub-flavored Markdown, you can make the list a checklist like this:

```markdown
- [x] user 1
- [ ] user 2
- [ ] ...
```

Might make sense to do for this.

garyb:
👍 from me

nwolverson:
This is fine.

csicar:
I am also fine with the license

andyarvanitis:
Fine for me, thanks

jdegoes:
👍

epost:
👍

paf31:
👍

kRITZCREEK:
I'm fine with the license.

stefanholzmueller:
Sounds good.

damncabbage:
I'm happy to license.

dgendill:
I'm fine with the license.

haf:
I'm fine with the license.

Xandaros:
Lol, I have changed a grand total of 1 line.
Yes, I'm fine with the license. Do what you want with my line ;)

kika:
[...]

PR #171, opened 2018-03-28
update pulp github url

juhp:


chexxor:
Thanks for the fix! We'll have to fix this link in [ecosystem/Editor-and-tool-support.md](https://github.com/purescript/documentation/blob/e2de1022d79ed980384ca31f90ca67484d775b07/ecosystem/Editor-and-tool-support.md) too, unless you add a patch with this PR.

Can you add a patch to include yourself in the Contributors list? https://github.com/purescript/documentation/blob/master/CONTRIBUTORS.md
I think we need that before we can merge.

juhp:
I can but it is such a small contribution, but if you really need it...

chexxor:
Yeah, I'm not an expert on copyright law, but we need explicit license to use the patches from all contributors if we want to redistribute or such.

It's motivated by the effort started in this issue: https://github.com/purescript/documentation/issues/151

hdgarrood:
It’s likely to be a lot easier in the long run to have a rule which everyone can agree on and which has no ambiguity, I think.

hdgarrood:
Everyone can agree on an interpretation of, that is.

juhp:
Sure of course I am happy to oblige - done!

(I would have been just as happy if someone had merged by hand :)

chexxor:
@juhp Thanks so much for helping improve the docs!

PR #181, opened 2018-05-18
Add docs for PartiallyAppliedSynonym

chexxor:
Type synonyms can be tricky to compose with data types, and likewise it's tricky to understand this type error. Here's a first draft for it.

### DIFF
--- a/errors/PartiallyAppliedSynonym.md
+++ b/errors/PartiallyAppliedSynonym.md
@@ -39,20 +39,23 @@ error = IdentityF (Const 5)
 
 ## Cause
 
-A type alias has been partially applied. To properly use a type alias having type parameters, like `type B a` or `type TypeIdentity a`, you must apply it to a type parameter at every location in the program, like `B Int` or `TypeIdentity Int`.
+A type synonym has been partially applied. To properly use a type synonym having type parameters, like `type B a` or `type TypeIdentity a`, you must apply it to a type parameter at every location in the program, like `B Int` or `TypeIdentity Int`.
 
-A type synonym *can* be partially applied in the context of a type synonym, however. For example:
+To understand the problem, it may be helpful to review the relevant difference between a "type synonym" and a "data type". A type synonym is effectively a macro, a substitution which is performed after parsing into an AST and before type-checking and compilation, while a data type is a proper entity through the compilation process. Because of this, a type synonym has much more restricted use, while a data type enjoys additional flexibility, such as the ability to be partially applied.
+
+Interestingly, a type synonym *can* be partially applied in the context of a type synonym, however. For example:
 
 ```purescript
 type UserT' f = { firstName :: f String }
-newtype UserNT' = UserNT' (UserT' TypeIdentity)
+u' :: UserT' TypeIdentity
+u' = { firstName: "test" }
 ```
 
 ## Fix
 
-Reconsider whether you want to be using a type alias for that particular type declaration. A type synonym is effectively a macro, a substitution which is performed after parsing into an AST and before type-checking and compilation, while a data type exists throughout the compilation process. Because of this, a type synonym has much more restricted use, while a data type enjoys the ability to be partially applied, like a function.
+Reconsider whether you want to use a type synonym for that particular type declaration, or whether to a data type is more appropriate.
 
-In the example above, if you want to use the type alias `TypeIdentity`, you will need to apply it to another type alias to produce a concrete type. That concrete type can then be used in a newtype or a function type signature.
+In the example above, if you want to use the type synonym `TypeIdentity`, you will need to apply it to another type synonym to produce a concrete type. That concrete type can then be used in a newtype or a function type signature.
 
 ```purescript
 type UserT' f = { firstName :: f String }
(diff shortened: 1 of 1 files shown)


errors/PartiallyAppliedSynonym.md
```diff
@@ -5,17 +5,33 @@
...
+type TypeIdentity t = t
+-- The following produces a `PartiallyAppliedSynonym` error.
+u :: UserNT TypeIdentity
+u = UserNT { firstName: "" }
 ```
 
 ## Cause
 
-Explain why a user might see this error.
+A type alias has been replaced with its value, and the result is that an expression has been partially applied.
```
garyb: This isn't right really, it's specifically to do with the synonym and the synonym alone:

``` purescript
data Const a b = Const a
data IdentityF f a = IdentityF (f a)

ok :: IdentityF (Const Int) Unit
ok = IdentityF (Const 5)

type A = Const

alsoOk :: IdentityF (A Int) Unit
alsoOk = IdentityF (Const 5)

type B a = Const a

alsoAlsoOk :: IdentityF (B Int) Unit
alsoAlsoOk = IdentityF (Const 5)

type X a b = Const a b

error :: IdentityF (X Int) Unit
error = IdentityF (Const 5)
```

Allowing partially applied type synonyms basically turns them into type level functions (an example here: https://stackoverflow.com/a/4923883), which sounds cool, but then it's not possible to check their equality in general.

errors/PartiallyAppliedSynonym.md
```diff
@@ -5,17 +5,33 @@
...
+
+In the example above, the data type `UserNT` is applied to the type alias `TypeIdentity t`. In the compiler, type aliases are replaced with the underlying type before type-checking, so the type-checker sees a type `UserNT` being applied to a type `t`.
 
 ## Fix
 
-- Suggest possible solutions.
+Reconsider how you're using type aliases and data types. Using a type aliase as if it is a normal data type may cause problems, as it not their intended purpose.
```
LiamGoodacre: `aliase` -> `alias`

errors/PartiallyAppliedSynonym.md
chexxor: > You say that this error is related to the type synonym alone. 

So this error has nothing to do with how it's replaced with its underlying type? But it's related to the underlying type somehow, too, no?

If the desugaring pass sees `IdentityF (X Int) Unit`, how would it know if the type synonym is partially applied? It would need to look at its underlying type, no?

Type synonyms seem benign and great... until you start looking at how to explain their interplay with data types. The problem seems to be a conflict with the syntax (what it appears to be) and the semantics (what it means). It looks like it can be used to do whatever type trickery you'd like, like you can with data types, but it's actually a trap.

To make a light metaphor, `data : Pikachu :: type : Mimikyu`: https://www.mymbuzz.com/wp-content/uploads/sites/2/2018/02/Detective-Pikachu-3DS-with-Mimikyu.jpg
It's a water-type! Mimikyu, electric attack! What? You can't do electric attacks? :)

I find this error to be absolutely unexplainable, which is so terribly frustrating. :(

I've looked at this error long enough and decided that it's unexplainable, sadly. The best docs I can come up with is to just not try to explain it and just give a code example of why it happens, and how to fix it.

I'll update.

chexxor: @garyb Oh! I see it now! I stared at your code snippet for ~20 minutes without getting any meaning from it. Your persistence wasn't for naught. I tried to explain it in my latest commit here.

In my explanation, there is still a thing which seems inconsistent. I'm fine with it, but I'm sure such an inconsistency will frustrate others.

> To properly use a type alias having type parameters, you must apply it to a type parameter at every location in the program...

and

> A type synonym *can* be partially applied in the context of a type synonym, however.

garyb: > A type synonym can be partially applied in the context of a type synonym, however.

Hmm, interesting! That's probably not intentional.

chexxor:
This is ready for a review again. Thanks so much @garyb !

garyb:
Might also be worth mentioning that when defining synonyms they should be defined as `un-applied` as possible, so they can be used in the widest range of situations (for example, the A/B/X types, the A version would be the most preferable definition).

(There's an issue to add a warning when defining synonyms like this so that the linter would suggest it anyway, but since we don't have it yet...)

chexxor:
I found this relatively long discussion on type synonyms: https://github.com/purescript/purescript/pull/2932#issuecomment-307657637
Lots of viewpoints discussed. I link from here for other people who come here and want to see more discussion.

chexxor:
@garyb I read through the type synonym discussions I could find in the purescript/purescript project, but I can't find anything which suggests `type A = Const` is better than `type B a = Const a`. Do you have a link to it?

I'll put that suggestion in here, but I need a better understanding of the situation. An existing discussion/related proposal would help me a lot here.

garyb:
There's not really much discussion about it because it's "obviously" worse when synonyms can be eta-reduced. Eta-expanding makes it impossible to use the synonym in some situations in which it would otherwise work, and unlike some other restrictions we impose on ourselves using the type system, this one never is never useful - it only prevents you doing things you might want to do, not things you might want to avoid.

An example, from transformers, if `State` was defined as:

``` purescript
-- note the `a`, in the library this is omitted on both sides
type State s a = StateT s Identity a
````
it would now be impossible to use `State` in some situations. When stacking transformers, only the top of the stack has its `a` provided: 
``` purescript
newtype App1 a = App1 (ExceptT MyError (State MyState) a)
newtype App2 a = App2 (WriterT MyLog (ExceptT MyError (State MyState)) a)
```
as the lower layers are used to fill in a type variable of kind `Type -> Type`. So in both the above cases, we're only applying `MyState` to `State`, meaning it is partially applied as we didn't include a value for the `a` in the synonym.

The issue for adding the warning is purescript/purescript#2691

chexxor:
Thanks @garyb! That's a great explanation! I've edited it a bit, to improve a reader's flow, and added it to the "notes" section in the error message.

It *might* be nice to include this to a section in "type synonym"s documentation.

PR #184, opened 2018-06-07
Update Getting Started for compiler v0.12.0

gabejohnson:
I did a fresh install of all tools as outlined in the guide and these updates will now allow a newcomer to complete the Euler task.

I included a workaround for `pulp init` which should be removed once https://github.com/purescript-contrib/pulp/pull/340 is merged.

I also added a note explaining that the book is currently a version behind.


guides/Getting-Started.md
```diff
@@ -44,6 +44,35 @@ You should see output similar to the following:
     * Build successful. Running tests...
     You should add some tests.
     * Tests OK.
+    
+If you instead see error messages containing `Module Control.Monad.Eff was not found.` this is likely because the code
+generated by `pulp init` is referencing the effect library widely used _before_ the release of the latest compiler (0.12.0) version.
+   
+To fix this error, change `src/Main.purs` and `test/Main.purs` to
```
paf31: I don't want to hold this PR up any more than necessary, but maybe we should solve this above by making sure they install and run the correct version of Pulp instead?

gabejohnson: I intend to remove this section as soon as a version of Pulp exists which generates the appropriate code 😄

PR #189, opened 2018-07-10
use Effect instead of Eff?

CarstenKoenig:
I guess it's better to just mention Effect instead of Eff, even if it does not provide the same granularity?

### DIFF
--- a/language/Differences-from-Haskell.md
+++ b/language/Differences-from-Haskell.md
@@ -70,7 +70,7 @@ Haskell uses the `IO` monad to deal with side effects. In PureScript, there is a
 main :: IO ()
 ```
 
-In PureScript the it's like this:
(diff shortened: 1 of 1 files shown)


chexxor:
Thanks for taking the time to fix this! <3

hdgarrood:
It would be better to remove that section entirely, actually. I don’t think the differences between `IO` and `Effect` are significant enough to be worth commenting on here.

CarstenKoenig:
@hdgarrood you are probably right - do you want me to add another PR?

hdgarrood:
Yes please, if you’re able. :)

hdgarrood:
Also can you please add your name to CONTRIBUTORS.md at the same time?

CarstenKoenig:
done - thanks

PR #190, opened 2018-07-11
Add hint for globstar usage in bash

Profpatsch:


hdgarrood:
That is not quite correct. The globs should be interpreted by the compiler, not the shell; the single quotes ought to ensure that the shell passes them through as is. We try to quote globs like this throughout docs to ensure that examples work regardless of shell configuration or indeed regardless of what OS you’re on (in particular, consider Windows).

Profpatsch:
Oh, I wasn’t aware the compiler itself interprets the globs!

PR #192, opened 2018-07-13
Remove mention of IO

CarstenKoenig:
as discussed here https://github.com/purescript/documentation/pull/189#issuecomment-404906508

there is now almost no difference between `IO` and `Effect` so it's not worth mentioning in this document


CarstenKoenig:
wait one moment - there is a mention of Eff there - I'll replace it in a minute

CarstenKoenig:
ok - I hope that's all

kRITZCREEK:
Looking good, thanks!

Issue #187, opened 2018-06-23
Errors in "Getting Started with Purescript"


senorsmile:
I'm trying out purescript for the first time.  

I'm attempting to follow the docs: 
https://github.com/purescript/documentation/blob/master/guides/Getting-Started.md

Things seem to install OK.  I can pulp init and successfully run tests.  

When I try to install purescript-lists, I get multiple options: 
```
bower install purescript-lists --save

Unable to find a suitable version for purescript-prelude, please choose one by typing one of the numbers below:
    1) purescript-prelude#3.3.0 which resolved to 3.3.0 and is required by install_test
    2) purescript-prelude#^3.0.0 which resolved to 3.3.0 and is required by purescript-eff#3.2.1
    3) purescript-prelude#^4.0.0 which resolved to 4.0.1 and is required by purescript-bifunctors#4.0.0, purescript-control#4.0.0, purescript-distributive#4.0.0, purescript-effect#2.0.0, purescript-either#4.0.0, purescript-foldable-traversable#4.0.0, purescript-identity#4.0.0, purescript-invariant#4.0.0, purescript-lazy#4.0.0, purescript-lists#5.0.0, purescript-maybe#4.0.0, purescript-newtype#3.0.0, purescript-nonempty#5.0.0, purescript-orders#4.0.0, purescript-refs#4.1.0, purescript-tailrec#4.0.0, purescript-tuples#5.0.0, purescript-unfoldable#4.0.0

```

There are a few more of these choices.  Regardless of which combo I pick, it always ends up in a state where running pulp repl errs out with:

```
pulp repl

Error found:
at bower_components/purescript-either/src/Data/Either/Inject.purs line 16, column 1 - line 16, column 1

  Unable to parse module:
  unexpected "else"
  expecting declaration or end of input


See https://github.com/purescript/documentation/blob/master/errors/ErrorParsingModule.md for more information,
or to contribute content related to this error.

```

Is this a problem or am I doing something very wrong?

justinwoo:
Seems like you have the wrong version of PureScript installed anyway, the current version is 0.12

senorsmile:
> Seems like you have the wrong version of PureScript installed anyway, the current version is 0.12

You are right.  I'm not too familiar with npm.  Simply following the docs ended up with 0.11.7.  Perhaps that's the source of the problem?

RamiroPastor:
I'm trying purescirpt for the first time too, and cannot even install it:

```
[...]

PR #199, opened 2018-08-11
Update Getting-Started.md

BebeSparkelSparkel:
Updated getting started for psc-package instead of bower

Caused me lots of frustration so hope this makes it easier for others.

Related issue #187

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -16,20 +16,21 @@ The Purescript compiler (`purs`) can be installed with npm:
 
 #### Setting up the Development Environment
 
-PureScript's core libraries are configured to use the [Pulp](https://github.com/purescript-contrib/pulp) build tool, and packages are available in the [Bower registry](http://bower.io/search/?q=purescript-).
+PureScript's core libraries are configured to use the [Pulp](https://github.com/purescript-contrib/pulp) build tool, and packages are installed with [psc-package](https://psc-package.readthedocs.io/). You can find the packages curated packages on [Pursuit](https://pursuit.purescript.org/)
 
-If you don't have Pulp and Bower installed, install them now:
+If you don't have Pulp and psc-package installed, install them now:
 
-    npm install -g pulp bower
+    npm install -g pulp psc-package-bin-simple
 
 Create a new project in an empty directory using `pulp init`:
 
-    pulp init
+    mkdir <package name>
+    cd <package name>
+    pulp --psc-package init
 
 Your directory should now contain the following files:
 
-- `bower.json` - contains library dependency information
-- `bower_components/` - a directory for installed dependencies
+- `psc-package.json` - contains library dependency information
 - `src/Main.purs` - Entry point module for your project
 - `test/Main.purs` - An empty test suite
 
@@ -49,9 +50,10 @@ If everything was built successfully, and the tests ran without problems, then t
 
 #### Installing Dependencies
 
-Dependencies can be installed using Bower. We will be using the `purescript-lists` library shortly, so install it now:
+Dependencies can be installed using psc-package. We will be using the `purescript-lists` library shortly, so install it now:
+*Note*: Package names on [Pursuit](https://pursuit.purescript.org/) and github all start with "purescript-" but when installing with psc-package remove that part of the package name.
 
-    bower install purescript-lists --save
+    psc-package install lists
 
 #### Working in PSCI
 


guides/Getting-Started.md
```diff
@@ -16,20 +16,21 @@ The Purescript compiler (`purs`) can be installed with npm:
 
 #### Setting up the Development Environment
 
-PureScript's core libraries are configured to use the [Pulp](https://github.com/purescript-contrib/pulp) build tool, and packages are available in the [Bower registry](http://bower.io/search/?q=purescript-).
+PureScript's core libraries are configured to use the [Pulp](https://github.com/purescript-contrib/pulp) build tool, and packages are installed with [psc-package](https://psc-package.readthedocs.io/). You can find the packages curated packages on [Pursuit](https://pursuit.purescript.org/)
```
justinwoo: >  You can find the packages curated packages on [Pursuit (https://pursuit.purescript.org/)

This seems confusing, that it would make people think that all Pursuit packages are in package sets. Also that package sets aren't actually curated, just community contributed.

BebeSparkelSparkel:
I guess my problem with bower was that I couldn't even install purescript-config and the default prelude for version 0.12.0 without having to manually/randomly pick the package versions

```text
Unable to find a suitable version for purescript-prelude, please choose one by typing one of the numbers below:
    1) purescript-prelude#^3.0.0 which resolved to 3.3.0 and is required by purescript-config#0.0.6, purescript-control#3.3.1, purescript-eff#3.2.1, purescript-either#3.2.0, purescript-functions#3.0.0, purescript-invariant#3.0.0, purescript-leibniz#4.1.0, purescript-maybe#3.1.0, purescript-newtype#2.0.0, purescript-node-streams#3.3.0, purescript-nonempty#4.3.0, purescript-posix-types#3.0.0, purescript-proxy#2.1.0, purescript-tuples#4.1.0
    2) purescript-prelude#^4.0.0 which resolved to 4.1.0 and is required by purescript-argonaut-core#4.0.1, purescript-arrays#5.0.0, purescript-bifunctors#4.0.0, purescript-console#4.1.0, purescript-const#4.0.0, purescript-contravariant#4.0.0, purescript-control#4.1.0, purescript-datetime#4.0.0, purescript-distributive#4.0.0, purescript-effect#2.0.0, purescript-either#4.0.0, purescript-enums#4.0.0, purescript-exceptions#4.0.0, purescript-foldable-traversable#4.0.0, purescript-foreign#5.0.0, purescript-foreign-object#1.0.0, purescript-form-urlencoded#4.0.1, purescript-functions#4.0.0, purescript-functors#3.0.1, purescript-gen#2.1.0, purescript-http-methods#4.0.2, purescript-identity#4.0.0, purescript-integers#4.0.0, purescript-invariant#4.1.0, purescript-lazy#4.0.0, purescript-lists#5.0.0, purescript-maybe#4.0.0, purescript-media-types#4.0.1, purescript-newtype#3.0.0, purescript-nonempty#5.0.0, purescript-ordered-collections#1.0.0, purescript-orders#4.0.0, purescript-parallel#4.0.0, purescript-psci-support#4.0.0, purescript-refs#4.1.0, purescript-st#4.0.0, purescript-strings#4.0.0, purescript-tailrec#4.0.0, purescript-transformers#4.1.0, purescript-tuples#5.0.0, purescript-unfoldable#4.0.0
    3) purescript-prelude#^4.1.0 which resolved to 4.1.0 and is required by rm2qb

Prefix the choice with ! to persist it to bower.json

? Answer 3
```

hdgarrood:
I accept that the UI could be a lot better in that case, but that is it working as expected; there does not appear to be a version of purescript-config which can work alongside version 4.1.0 of the prelude. Therefore switching to psc-package wouldn't help you there: purescript-config most likely wouldn't even make it into any 0.12.x package set at all in its current state.

BebeSparkelSparkel:
So there is no way to access env vars right now without downgrading to 0.11.7

hdgarrood:
That’s not true, see https://pursuit.purescript.org/packages/purescript-node-process/6.0.0/docs/Node.Process#v:lookupEnv

BebeSparkelSparkel:
Thanks

Is https://purescript-users.ml/t/working-with-0-11-7-in-a-0-12-world/180 still valid or is it out of date at this point?

hdgarrood:
A lot more of the ecosystem has caught up by this point in comparison to when that was written. I think I’d probably recommend 0.12 unless you have a specific reason to use 0.11.7.

hdgarrood:
I think switching the getting started guide to use psc-package rather than bower is potentially worth considering but needs more discussion first -- for instance, I'd argue we'd also need to a) explain the differences between the psc-package model vs the more standard centralised package repo + dependency solving approach which bower (and most other package managers take) and b) decide on and commit to a standard way of shipping the `psc-package` binary. Suggest closing this (and potentially also opening a new issue for this discussion)?

JordanMartinez:
> I'd argue we'd also need to a) explain the differences between the psc-package model vs the more standard centralised package repo + dependency solving approach which bower (and most other package managers take)

@hdgarrood I've already done most of this (aside from the dependency solving approach Bower uses) in [the related page in my learning repo](https://github.com/JordanMartinez/purescript-jordans-reference/blob/latestRelease/01-Build-Tools/01-Tool-Comparisons/01-Dependency-Managers.md)

hdgarrood:
@JordanMartinez That page looks good but on reflection I'm not sure it's appropriate to expose all that to someone who is just trying to get something working during their first experience with the language. I'm going to close this PR for now, but feel free to open issues if you want to revive this discussion.

PR #202, opened 2018-08-31
add documentation on Instance Chains

justinwoo:


language/Type-Classes.md
```diff
@@ -25,6 +23,36 @@ instance showArray :: (Show a) => Show (Array a) where
 example = show [true, false]
 ```
 
+Overlapping instances are no longer allowed in PureScript. To write overlapping instances, you should use Instance Chains.
+
+## Instance Chains
+
+PureScript current implements a form of instance chains that work on groups of instances matching by parameters. This means that constraints are not considered when choosing instances. However, you can still write a chain of instances in consecutive order that will be matched top to bottom by use the `else` keyword.
```
LiamGoodacre: Typos:
* `PureScript current implements` -> `PureScript currently implements`
* `by use the` -> `by using the`

justinwoo: Fixed, thanks

chexxor:
LGTM

PR #211, opened 2018-09-15
fill in Multi-Param Type Classes

justinwoo:


language/Type-Classes.md
```diff
@@ -55,7 +55,29 @@ main = do
...
+else instance injectRight :: Inject a b => Inject a (Either c b) where
+  inj = Right <<< inj
+  prj = either (const Nothing) prj
+
+else instance injectReflexive :: Inject a a where
+  inj = identity
+  prj = Just
+```
+
+And our instances will be matched when both type parameters are available. See the section on Functional Dependencies for cases hwen one or more parameters may be determined by another.
```
LiamGoodacre: `hwen` -> `when`

justinwoo: oops, fixed

language/Type-Classes.md
```diff
@@ -55,7 +55,29 @@ main = do
...
+
+instance injectLeft :: Inject a (Either a b) where
+  inj = Left
+  prj = either Just (const Nothing)
+
+else instance injectRight :: Inject a b => Inject a (Either c b) where
+  inj = Right <<< inj
+  prj = either (const Nothing) prj
+
+else instance injectReflexive :: Inject a a where
```
LiamGoodacre: Shouldn't this (`injectReflexive`) be the first instance?
Suppose `a` is `Either Int String`, currently this would hit the `injectRight` case and then try `Inject (Either Int String) String` - which isn't quite what you might expect.

justinwoo: Example taken from https://github.com/purescript/purescript-either/blob/v4.0.0/src/Data/Either/Inject.purs#L8-L10

chexxor: Maybe we can put this link below the example? This would increase the chance that a reader will notice this example differs from its source, should the source ever be changed/bug-fixed.

Would be nice to avoid making you come up with a whole-new example.

hdgarrood: Shall we resolve this in the `purescript-either` repo first? https://github.com/purescript/purescript-either/issues/43

hdgarrood:
We've resolved the issue in the `purescript-either` repo, but on second thoughts, I think this example is probably not best suited because of the use of instance chains, which I would consider to be something of a 'power user' feature. What about, say, `MonadState` instead?

justinwoo:
I was thinking of something simpler along the lines of IsTypeEqual

hdgarrood:
Yeah, that's a good option, although another problem with both of these is that the functional dependencies could also risk obscuring what we're trying to get across. Are there even any good example of MPTCs with no functional dependencies?

natefaubion:
I think it would be good to use something like `MonadAsk` both with and without functional dependencies.

natefaubion:
`MonadError` is another good one. I think these are better than `MonadState` because the case of wanting to project different environments or throw different errors comes up more often for me than modifying multiple state.

language/Type-Classes.md
chexxor: That issue has been resolved, it appears: https://github.com/purescript/purescript-either/issues/43

justinwoo:
Cleaning up my github notifications

PR #215, opened 2018-10-19
Update Editor-and-tool-support.md

justinwoo:
Does it even make sense to say "experimental" when Psc-Package has been around for a long while and works fairly well? I use it for work and almost everything else.


LiamGoodacre:
Sounds good to me.

PR #217, opened 2018-10-31
Add page with FAQs

ad-si:


hdgarrood:
It looks to me like the existing  ‘differences from Haskell’ page would be a better place for this information (if it’s not already on there).

language/FAQ.md
```diff
@@ -0,0 +1,36 @@
...
+
+Using `=` would lead to ambiguities:
+
+Does `book {title = "Hello"}` mean "update the title of book to `Hello`",
+or does it mean "apply the function `book` to the record `{title = "Hello"}`".
+
+Using `:` avoids this ambiguity.
+
+
+## Why is `>>>` instead of `.` used for function composition?
```
LiamGoodacre: I think the `>>>` here should be `<<<` which is a composition that is more similar to `.`.

hdgarrood:
Thanks, but I'm going to close this. I'm not convinced by the SEO argument, because all of the questions in this file already have the relevant threads come up as the first answer when you google them (admittedly, apart from "Why is `:` instead of `=` used for record assignments?" but answering questions like that isn't really what this repo is for). Also, the "Differences from Haskell" page does already explain why instances must be named and why we don't use `.` for composition. Admittedly the explanation of why instances must be named isn't as in-depth as the one in this PR, but I don't think a hugely in-depth explanation would be appropriate there anyway.

Issue #216, opened 2018-10-19
Add link to "purescript-jordans-reference" repo


JordanMartinez:
Per @kRITZCREEK's response in purescript/purescript.github.io#132

Link to my learning repo found here:
https://github.com/JordanMartinez/purescript-jordans-reference

PR #220, opened 2018-11-28
Add link to purescript-jordans-reference in Learning section

JordanMartinez:
I opened #216 but never got a response. So, I'm trying this approach instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Feel free to make an issue to discuss amending the scope.
 
 - The [PureScript Book](https://leanpub.com/purescript/read) is the recommended approach to learning the language, since it covers more material in greater depth. However, it is not updated yet for the 0.12 version of the compiler.
 - [Language Reference](language/README.md)
-- [PureScript: Jordan's Reference](https://github.com/JordanMartinez/purescript-jordans-reference): An up-to-date project covering Getting Started, Build Tools, PureScript's verified-by-compilation syntax via meta-language, FP design patterns, and PureScript's ecosystem.
+- [PureScript: Jordan's Reference](https://github.com/JordanMartinez/purescript-jordans-reference): An up-to-date project covering Getting Started, Build Tools, PureScript's syntax with examples, FP design patterns, and PureScript's ecosystem.
 
 ## Guides
 
@@ -78,4 +78,3 @@ Feel free to make an issue to discuss amending the scope.
 
 - [Related Projects](Related-Projects.md)
 - [Differences from Haskell](language/Differences-from-Haskell.md)
-
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Feel free to make an issue to discuss amending the scope.
 
 - The [PureScript Book](https://leanpub.com/purescript/read) is the recommended approach to learning the language, since it covers more material in greater depth. However, it is not updated yet for the 0.12 version of the compiler.
 - [Language Reference](language/README.md)
+- [Purescript: Jordan's Reference](https://github.com/JordanMartinez/purescript-jordans-reference): An up-to-date project covering Getting Started, Build Tools, Purescript's verified-by-compilation syntax via meta-language, FP design patterns, and PureScript's ecosystem.
 
 ## Guides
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,6 +35,7 @@ Feel free to make an issue to discuss amending the scope.
 
 - The [PureScript Book](https://leanpub.com/purescript/read) is the recommended approach to learning the language, since it covers more material in greater depth. However, it is not updated yet for the 0.12 version of the compiler.
 - [Language Reference](language/README.md)
+- [PureScript: Jordan's Reference](https://github.com/JordanMartinez/purescript-jordans-reference): An up-to-date project covering Getting Started, Build Tools, PureScript's verified-by-compilation syntax via meta-language, FP design patterns, and PureScript's ecosystem.
```
chexxor: What is "verified-by-compilation syntax via meta-language"? I'd like to use more common words for this idea, if possible.

hdgarrood:
Sorry for the delay in responding. The reason it's taken a while is that before linking to things, I'd really want to make sure we've had a good look at them first, and your repo has a decent amount of stuff in it.

I'm happy to put links to unofficial documentation resources in here, but I'd prefer that we clearly separated them from any official documentation and added a note to clarify that they are unofficial.

> Content from other people's projects can be referenced by contributors to this project, when and if that happens.

I'm not super keen on this, as it happens; we have enough trouble keeping links between official resources alive as it is, and any references to unofficial resources are even more likely to change under us.

chexxor:
>  I'm not super keen on this, as it happens; we have enough trouble keeping links between official resources alive as it is, and any references to unofficial resources are even more likely to change under us.

I should have clarified that - I mean that when people are aiming to write new documentation for this purescript/documentation project, they can browse through other people's doc projects to find docs to start from.

README.md
JordanMartinez: :sweat_smile: Yeah, that's a good point.

Basically "meta-language" means I use the name of the actual terms to explain what they represent:
```purescript
-- instead of:
f :: Int -> String
f x = show x

-- I write:
function :: ArgType -> ReturnType
function argValue = show argValue

-- and again
class TypeClass someType where
  someFunction :: someType -> String

-- needed to compile
type ArgType = Int
type ReturnType = String
```
"Verified-by-compilation" means the above syntax is actual Purescript code that compiles. So, rather than me saying "Hey! I claim that this is correct; just trust me!", I can say "Hey, I can prove that this works; no need to trust me." (Obviously, there's no CI that backs up that claim yet, but it's on my todo list).

JordanMartinez:
One thought that's also been on my mind. Since my repo is "unofficial", what if I just contributed a large portion of it to this repo? There might be some maintenance concerns, but it's a thought that's been on my mind.

README.md
JordanMartinez: As an example of the meta-language, see [how I explained Pattern Matching](https://github.com/JordanMartinez/purescript-jordans-reference/blob/latestRelease/11-Syntax/01-Basic-Syntax/src/02-Data-and-Functions/05-Pattern-Matching-in-Functions.purs)

chexxor:
I think the maintainers of the purescript/documentation project want to keep the scope relatively manageable. I think this means compiler docs and some core tools docs -- mostly just reference documentation, not walkthroughs or how to FP design patterns.

JordanMartinez:
Good points!

README.md
chexxor: Hmm, I would just say "...covering "Getting Started", build tools, syntax with examples, FP design patterns, and the ecosystem. If you decide to change how you explain it, or what you call it, we wouldn't have to make a patch to this purescript/documentation project, also.
What do you think? Do you feel strongly about your original choice of words?

chexxor:
@JordanMartinez Is it still on your radar to add a patch to address the comment I made on your addition?

JordanMartinez:
Oh, I thought we were still determining what the description of the link should say. So, I was waiting for others to say "how about 'x'" or something.

Since it's been a while since I've looked at this, what needs to be changed/fixed for this to be merged?

README.md
chexxor: Up here, I'd like this language to be more simple.

JordanMartinez:
Let me know if there's anything else that needs to be done.

chexxor:
LGTM - I'll merge in several hours, to give time for comments.

PR #235, opened 2019-01-24
update pattern matching documentation to show multiple expressions

justinwoo:


language/Pattern-Matching.md
```diff
@@ -176,6 +176,17 @@ compare _ _ = EQ
...
+
+```purescript
+positiveLessThanFive :: Maybe Int -> Boolean
+positiveLessThanFive mInt
+  | Just x <- mInt
+  , x > 0
+  , x < 5 = true
+  | otherwise = false
+```
+
```
natefaubion: Should this be moved into the `Pattern Guards` section below? Commas are only necessary because of pattern guards.

justinwoo: oh yeah, i'll move it down

PR #242, opened 2019-02-23
Update MissingTypeDeclaration.md

Beyarz:
Filled in with an example and solution.

### DIFF
--- a/errors/MissingTypeDeclaration.md
+++ b/errors/MissingTypeDeclaration.md
@@ -29,4 +29,5 @@ main = log "The type is now declared!"
 
 ## Notes
 
-- Additional notes.
+- MissingTypeDeclaration is not an error, it is a warning.
+- Type annotations are necessary in cases where you're using type classes in an ambiguous way.
--- a/errors/MissingTypeDeclaration.md
+++ b/errors/MissingTypeDeclaration.md
@@ -3,9 +3,11 @@
 ## Example
 
 ```purescript
-module ShortFailingExample where
+module Main where
+import Prelude
+import Effect.Console (log)
 
-...
+main = log "Main type is undefined."
 ```
 
 ## Cause
@@ -14,7 +16,16 @@ Explain why a user might see this error.
 
 ## Fix
 
-- Suggest possible solutions.
+- Declare the type
+```purescript
+module Main where
+import Prelude
+import Effect (Effect)
+import Effect.Console (log)
+
+main :: Effect Unit
+main = log "The type is now declared!"
+```
 
 ## Notes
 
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Does one always need to give the type declaration? I thought one can remove it in some situations. 

In your example, it is required because it's the entry point into the program. For others, I think their type signature can be inferred by the context in which they are used and their body's definition.

hdgarrood:
Type annotations are only necessary in rare situations where you have rank-2 or higher types (like https://pursuit.purescript.org/packages/purescript-st/4.0.0/docs/Control.Monad.ST#v:run). It's not necessary in this example. In this case it's only a warning, but warnings and errors are both kept in this directory.

errors/MissingTypeDeclaration.md
```diff
@@ -3,9 +3,11 @@
 ## Example
 
 ```purescript
-module ShortFailingExample where
+module Main where
+import Prelude
+import Effect.Console (log)
 
-...
+main = log "Main type is undefined."
```
hdgarrood: "undefined" isn't quite the right word to use here. The type is defined; it's inferrable, it can only ever be `Effect Unit`. It's just not written explicitly in the source file. How about "No type annotation has been provided." ?

hdgarrood:
> Type annotations are only necessary in rare situations where you have rank-2 or higher types

Sorry, this isn't quite true: they're also necessary in cases where you're using type classes in an ambiguous way, e.g.

```purescript
roundtrip :: Json -> Either String Json
roundtrip = decodeJson >>> map encodeJson
```

errors/MissingTypeDeclaration.md
Beyarz: Sounds good to me!

Beyarz:
> > Type annotations are only necessary in rare situations where you have rank-2 or higher types
> Sorry, this isn't quite true: they're also necessary in cases where you're using type classes in an ambiguous way, e.g.
I added these as notes

Beyarz:
> Does one always need to give the type declaration?

Not it is not always needed, I clarified that in the notes.

Beyarz:
How could we explain the MissingTypeDeclaration warning?

Beyarz:
> In your example, it is required because it's the entry point into the program

I changed from Main to ShortFailingExample,

Beyarz:
- New module name

Beyarz:
I would like to fill the cause of the warning with some more details but I can't think of anything more to add really. Not at the moment.

hdgarrood:
> I would like to fill the section explaining the cause of the warning with a bit more details but I can't think of anything more to add really. Not at the moment.

No worries, I'll add a bit more to fill it in. @Beyarz can you add yourself to CONTRIBUTORS.md please?

Beyarz:
> No worries, I'll add a bit more to fill it in.
Thanks!

>@Beyarz can you add yourself to CONTRIBUTORS.md please?
Of course!

hdgarrood:
Thanks!

PR #245, opened 2019-03-03
fixed wrong type signature

amerocu:


hdgarrood:
Thanks!

bernhard-herzog:
ISTM that this original type was correct. Before this change, the example in question read:

```purescript
-- Create an alias for a complex function type
type Baz = Number -> Number -> Bar Number

-- This function will take two arguments and return a record with double the value
mkDoubledFoo :: Baz
mkDoubledFoo foo bar = { foo: 2.0*foo, bar: 2.0*bar }
```

`Baz` is defined as a type alias immediately before `mkDoubledFoo` and is the correct type for that function (`Bar` is defined earlier in a part I didn't quote here). With this pull request the type becomes `Number -> Number -> Baz` which is clearly wrong.

hdgarrood:
:man_facepalming: thank you. If someone wants to put a PR together reverting the change I’d appreciate it!

bernhard-herzog:
I'll prepare one

PR #247, opened 2019-03-12
Note difference in behavior of range operator (..)

bernhard-herzog:
In particular note that the behavior of (a..b) differs from [a..b] when
b is smaller than a.

This difference in behavior has cost a couple hours last week when I had to debug a piece of code I had ported from Haskell.


hdgarrood:
:+1: Thanks!

PR #249, opened 2019-03-26
Chexxor rows in instances

RathishB23:
Add data type examples for a documentation project


hdgarrood:
The comment doesn’t match up with the commits in this PR and it looks like it is basically the same as #104, so I’m going to assume this was opened in error.

JordanMartinez:
Similar think happened to me in JordanMartinez/purescript-jordans-reference#301

I'm not sure why that happened.

PR #252, opened 2019-05-26
Fix: broken link in MissingTypeDeclaration.md

newswim:
👋 

Just a tiny little change, I think the relative link was broken due to a missing prepended `/`.


hdgarrood:
Thanks!

PR #255, opened 2019-06-27
Update editor and tools section

drewolson:
Add `w0rp/ale` to the vim section and `spago` to the build tools section.


hdgarrood:
Thanks!

PR #260, opened 2019-07-21
Add section on multiple expression case matching

### LINKED ISSUE #259: Mention multi-case patterns in "Pattern Matching"

alextes:
I think the example could be a bit more abstract and straightforward 🤔 . Let me know what you think.

My understanding is also that this applies to case expressions specifically, which is why I nested it there. Let me know if it should instead have a separate pattern section.

Fixes #259

### DIFF
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -35,6 +35,13 @@ The following forms can be used for matching:
 
 Guards and pattern guards are also supported.
 
+One can also match multiple expressions at once.
+```purescript
+case maybeName, UserStatus of
+  _, NotFound -> doA
+  (Just name), Exists -> doB name
+```
+
 The exhaustivity checker will introduce a `Partial` constraint for any pattern which is not exhaustive.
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think it would probably be better to put this section at the end of this file, because I think it relies on the reader already having a basic grasp of patterns. Also I think it might be good if the example was a little more self-contained, although I don't have any more specific suggestions off the top of my head right now

hdgarrood:
I just realised that this is already documented in https://github.com/purescript/documentation/blob/master/language/Syntax.md#case-expressions

alextes:
Hm, shall we add a link to that section maybe?
I don't think I'd be quick to search for syntax. My search was: "[tuple pattern match purescript](https://www.google.nl/search?q=tuple+pattern+match+purescript)," but even noting that I was falling for the XY Problem there and looking in more general terms for "[multiple expression case match purescript](https://www.google.nl/search?q=multiple+expression+case+match+purescript)," it's the second result.

I would say that's good enough and close the issue with that. If there are proponents for adding a link in the section that is admittedly about pattern matching, that makes sense to me too.

alextes:
It sounds like there are no proponents. Closing 😊!

PR #261, opened 2019-07-27
record update function

BebeSparkelSparkel:
https://github.com/purescript/purescript/issues/3704


hdgarrood:
I think it would be good to mention this here but I don't think we should bring monadic state into it; I think an example which just defines a function using this syntax and then applies it to a record should be sufficient.

hdgarrood:
Thanks!

PR #275, opened 2019-10-01
Calling functions with constrained types from JS is unsafe

### LINKED ISSUE #273: FFI Guide shouldn't talk about passing dictionaries on the JS side
We generally consider passing dictionaries as an implementation detail that users shouldn't rely on for their FFI as specialization optimizations would break FFI code written like that.

The guide in here https://github.com/purescript/documentation/blob/2ecc0e6bea10ecd0304252363b574af309ca59d4/guides/FFI.md#handling-constrained-types talks about how to do that though, so we should probably get rid of that section, or talk about how to explicitly pass the functions you need from the dictionary into the FFI similar to how we do it for data constructors.

kritzcreek:
Fixes #273

### DIFF
--- a/guides/FFI.md
+++ b/guides/FFI.md
@@ -156,45 +156,11 @@ This time, we can assign the curried function type directly:
 foreign import calculateInterest :: Number -> Number -> Number
 ```
 
-#### Handling Constrained Types
+#### Don't call functions with constrained types from JavaScript
 
-Another special case that you should be aware of when calling PureScript functions from Javascript is that values with constrained types (i.e. types which contain type class constraints) contain extra parameters which are used to pass type class dictionaries to the function.
+It's not safe to call functions that have typeclass constraints from JavaScript, as their representation is unstable in the face of compiler optimizations. If you want to call a PureScript function from JavaScript you need to take all arguments explicitly.
 
-For example, let's write a simple PureScript function with a constrained type, and look at the generated Javascript.
-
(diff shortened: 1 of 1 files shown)


guides/FFI.md
```diff
@@ -156,45 +156,11 @@ This time, we can assign the curried function type directly:
 foreign import calculateInterest :: Number -> Number -> Number
 ```
 
-#### Handling Constrained Types
+#### Don't call functions with constrained types from JavaScript
 
-Another special case that you should be aware of when calling PureScript functions from Javascript is that values with constrained types (i.e. types which contain type class constraints) contain extra parameters which are used to pass type class dictionaries to the function.
+It's not safe to call functions that have typeclass constraints from JavaScript, as their representation is unstable in the face of compiler optimizations. If you want to call a PureScript function from JavaScript you need to take all arguments explicitly.
```
natefaubion: I think you could clarify this a little. All arguments on the JS side are explicit, but the point is that some arguments would have been filled in by the compiler. I think you could clarify that you can export a function for use in JS with a monomorphic signature which will result in the compiler pre-applying the dictionary.

kritzcreek: Thanks for the feedback, I struggled with expressing exactly what you're saying here. I'll sit down and try to formulate it.

kritzcreek:
I'll close this as I don't have the time to get to this.

PR #279, opened 2019-10-14
Remove duplicate PureScript by Example book link

milesfrain:
New users may see `PureScript By Example` and `PureScript Book` as two separate resources.

Even if these links are renamed to match, new users could still miss the warning in **Learning** if they dive straight into **Getting Started**.

It may be best to just remove `PureScript By Example` from the getting started section, or merge both sections.


hdgarrood:
Thanks! I think removing it from the Getting Started section is a good idea.

milesfrain:
The `Getting Started` content could also be merged with `PureScript By Example, Chapter 2, Getting Started`, but we can save that for another PR.

hdgarrood:
Could you add yourself to CONTRIBUTORS.md please?

hdgarrood:
I think it's best not to try to merge anything in this repo with what's in the book, by the way. I think getting the compiler running deserves its own standalone page, which we can update as required without necessarily needing to wait for Phil's input.

milesfrain:
Also posted a link to the most complete set of exercise solutions I could find.
I'd like to eventually update these exercise solutions and create another "unsolved" repo with unit tests than new learners can use to check their solutions.

hdgarrood:
Actually, I'd rather not link to solutions if they haven't been looked over by a maintainer, and especially if they are for an older compiler version.

hdgarrood:
Can we just do the deduplication of the purescript book part in this PR, and discuss the other changes separately please?

milesfrain:
Here's the minimal change to remove the duplicate link.

Also setup #280 with the other changes.

hdgarrood:
Thanks!

PR #280, opened 2019-10-20
Link to latest Purescript By Example resources

milesfrain:
This change adds a link to the most recent update effort of the purescript book along with the only complete set of exercise solutions.
These resources are not yet in a perfect state, but I suspect new users would still benefit from them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,12 @@ This repository is a collaborative effort, so please feel free to make a pull re
 
 ### Learning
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
Thanks, but I don't think we should add all of these extra links here; there's a risk of overloading a new user with too much choice. I also would rather not link to solutions here; the value of an exercise comes from thinking about how to solve it - the solution itself is just a byproduct. There are also plenty of avenues for people to get help for a particular exercise if they happen to be stuck (eg Slack or Stack Overflow).

milesfrain:
I agree with the overloading concern. Hopefully the dedup change will let future new users find these links more easily than I did. I'll close this PR.

I'd also trust folks to only use the solutions to check their own work, but perhaps the temptation will be too much and encourage giving-up earlier.

PR #281, opened 2019-10-31
Documented declConflict errors

Woody88:
I am not sure if there are other cases when this error occurs, but I think that this should be fine? I am pretty sure that there is a better way to explain this, but this is what I came up with quickly.

### DIFF
--- a/errors/DeclConflict.md
+++ b/errors/DeclConflict.md
@@ -3,19 +3,28 @@
 ## Example
 
 ```purescript
-module ShortFailingExample where
+data MyType = A | B 
 
-...
+data MyType2 = C | D
+
+data MyType3 = MyType3 { x :: Int }
+
+data MyType4 = MyType | MyType2 | MyType3 
 ```
 
 ## Cause
 
-Explain why a user might see this error.
+This error shows up when an a defined type (i.e: MyType, MyType2, etc...) is being defined as a data constructor for another data type definition (i.e: MyType4). 
 
 ## Fix
+This can be fixed by simply wrapping these pre-defined data types in a new data constrcutor.
+```purescript
+data MyType = A | B 
 
-- Suggest possible solutions.
+data MyType2 = C | D
 
-## Notes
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think this is a little overly specific, as there are indeed a few other cases where this can show up. For example, repeated data constructors, or data constructors whose names conflict with type class names. See `tests/purs/failing/DeclConflict*.purs` in the compiler repo for more examples.

I don't think your suggested fix really works, though, because writing, say, `data X = Y` is very different from writing `data X = X Y`. The former is a type with precisely one inhabitant which you've called `Y`, the latter is a type which has the same number of inhabitants as some other existing type `Y`. I think the suggested fix should be to either come up with new names which don't conflict, or to put the conflicting names in a separate module.

hdgarrood:
Just to be clear, this file doesn't have to exhaustively list the possibilities for an error like this to occur; I think that would probably end up being counterproductive. I think we should just provide a couple of very basic examples, around the minimum size required to produce this error.

thomashoneyman:
@Woody88 what do you think of an update to your PR to something like this?

---

# `DeclConflict` Error

## Example

Each of these pairs will fail with a `DeclConflict` error.

```purescript
module Main where

data T = Fail | Fail

data T1 = Fail1
data T2 = Fail1

class Fail2
data T3 = Fail2
```

## Cause

This error occurs when a data constructor, type, or type class conflicts with another data constructor, type, or type class.

## Fix

This can be fixed by using non-conflicting names or by putting conflicting names into a separate module. Then, if the names are later used together in a module they can be distinguished via a qualified import:

```purs
module Main where

import Module1 as Module1 -- includes: data Works = Works
import Module2 as Module2 -- includes: data Works = Works

x :: Module1.Works
x = Module1.Works

y :: Module2.Works
y = Module2.Works
```

Woody88:
Hi @thomashoneyman @hdgarrood! 

No problem, I will add the proposed changes tomorrow.

hdgarrood:
Thanks!

PR #286, opened 2020-01-29
Update getting started guide to use spago

drewolson:
This updates the existing guide as-is to use `spago` rather than `pulp` and `bower`. There's probably more clean up to be done here, but I wanted to keep this simple.

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -46,7 +46,7 @@ You should see output similar to the following:
     You should add some tests.
     [info] Tests succeeded.
 
-If everything was built successfully, and the tests ran without problems, then the last line should state "Tests OK".
+If everything was built successfully, and the tests ran without problems, then the last line should state "Tests succeeded."
 
 #### Installing Dependencies
 
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -2,7 +2,7 @@
 
 Let's walk through the basics of getting set up to use the PureScript compiler `purs`, and its interactive mode `purs repl`.
 
-We'll start with the installation of the compiler and Pulp build tool, and then go through the basic usage of `purs repl`, working towards a solution of problem 1 from [Project Euler](http://projecteuler.net/problem=1).
+We'll start with the installation of the compiler and Spago build tool, and then go through the basic usage of `purs repl`, working towards a solution of problem 1 from [Project Euler](http://projecteuler.net/problem=1).
 
 #### Installing the Compiler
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
I'd really rather get said cleanup done up front before merging in this case, as this is basically people's first experience with the language. Can you run through the guide as if you were a beginner following it, to make sure everything works the same way?

hdgarrood:
cc @f-f would you mind taking a look at this?

guides/Getting-Started.md
```diff
@@ -89,10 +89,11 @@ We will use a selection of these commands during this tutorial.
 
 Start by pressing the Tab key to use the autocompletion feature. You will see a collection of names of functions from the Prelude which are available to use.
 
-To see the type of one of these values, first import the appropriate module using the `import` command. `pulp init` configures `.purs-repl` to install `Prelude` automatically, so you won't have to do it yourself.
```
hdgarrood: @f-f what do you think about having Spago create a `.purs-repl` file including `import Prelude`, like Pulp does, so that the repl "just works"?

hdgarrood: Having it create this file during spago init, I mean

hdgarrood:
> Can you run through the guide as if you were a beginner following it, to make sure everything works the same way?

Oh sorry, it looks like you already did this. Never mind :sweat_smile: what cleanup did you have in mind?

drewolson:
@hdgarrood just overall structure and tone, less "clean up" and more "rethinking".

But yes, just to be explicit, I did run through the entire guide in my console to make sure all commands work.

guides/Getting-Started.md
```diff
@@ -16,48 +16,48 @@ The Purescript compiler (`purs`) can be installed with npm:
 
 #### Setting up the Development Environment
 
-PureScript's core libraries are configured to use the [Pulp](https://github.com/purescript-contrib/pulp) build tool, and packages are available in the [Bower registry](http://bower.io/search/?q=purescript-).
+PureScript's core libraries are configured to use the [Spago](https://github.com/spacchetti/spago) package manager and build tool.
```
thomashoneyman: If the move to `purescript/spago` will happen quite soon we need to make sure to update this link (though fortunately GitHub will redirect automatically in the worst case).

drewolson: Yes, I was thinking about this as I pasted the link. That said, the update should be speedy (and, as you said, the redirect will work in the interim).

guides/Getting-Started.md
```diff
@@ -16,48 +16,48 @@ The Purescript compiler (`purs`) can be installed with npm:
...
     You should add some tests.
     * Tests OK.
-    
+
 If everything was built successfully, and the tests ran without problems, then the last line should state "Tests OK".
 
 #### Installing Dependencies
 
-Dependencies can be installed using Bower. We will be using the `purescript-lists` library shortly, so install it now:
+Dependencies can be installed using Spago. We will be using the `purescript-lists` library shortly, so install it now:
```
thomashoneyman: Is it worth adding a line here that by convention / for historical reasons PureScript packages have a `purescript-` prefix, but when installing them you just use the package name? 

I can see it being a little odd to instruct someone to install `purescript-lists` by typing `spago install lists` without them knowing whether that's how all packages work or just this one.

Fortunately I believe Spago will install `purescript-lists` by trimming the `purescript-` prefix.

hdgarrood:
Yes, I suppose that's true. Ok, how about I'll merge this now, create an issue for remaining Bower updates (there are one or two other places in this repo we should probably update), and leave the rethink idea open for now; maybe we can wait and see if anything in here as it is now trips beginners up.

JordanMartinez:
Just fyi. Let me know whether your want to use any of the content in my
learning repo.

>

guides/Getting-Started.md
f-f: > Fortunately I believe Spago will install `purescript-lists` by trimming the `purescript-` prefix.

This is right. I think it's ok to reference the package with the prefix, as that's the name of the repo anyways? Though I have no strong opinion on this :man_shrugging:

guides/Getting-Started.md
f-f: @hdgarrood yeah this is a great idea! I opened https://github.com/spacchetti/spago/issues/555 to track this

guides/Getting-Started.md
```diff
@@ -16,48 +16,48 @@ The Purescript compiler (`purs`) can be installed with npm:
...
 
-    pulp build
-    pulp test
+    spago build
+    spago test
 
 You should see output similar to the following:
 
-    * Building project in /Users/paf31/Documents/Code/purescript/pulp-test
+    * Building project in /Users/paf31/Documents/Code/purescript/spago-test
```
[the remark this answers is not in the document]
drewolson: Fixed in 4d921e41be77c5e87d3b958e983184223da08e7e

guides/Getting-Started.md
```diff
@@ -40,10 +40,11 @@ At this point, you should be able to build the project and run the tests:
...
-    * Building project in /Users/paf31/Documents/Code/purescript/spago-test
-    * Build successful. Running tests...
+    [info] Installation complete.
+    [info] Build succeeded.
+    🍝
     You should add some tests.
-    * Tests OK.
+    [info] Tests succeeded.
 
 If everything was built successfully, and the tests ran without problems, then the last line should state "Tests OK".
```
thomashoneyman:
```suggestion
If everything was built successfully, and the tests ran without problems, then the last line should state "Tests succeeded."
```

drewolson:
Is there any more feedback preventing this from being merged? No rush, just wanted to make sure I've addressed everything.

thomashoneyman:
I don't have write access to this repository, but this looks good to me and I will merge the accompanying PR https://github.com/purescript/purescript.github.io/pull/138 tonight if there are no objections (or no one merges it before me 🥇)

hdgarrood:
Thanks!

PR #292, opened 2020-02-16
Add a link to Editor-and-tool-support.md

kevinbarabash:
I noticed there was a link missing in the ecosystem section.


hdgarrood:
Actually never mind, I just spotted that it is already linked under "tools", which is probably more appropriate. Thanks anyway though!

PR #295, opened 2020-02-29
Replace outdated references to Pulp with Spago in PSCi Guide

mbbx6spp:
- Replaces references of `pulp` with `spago`
- Adds step to install `psci-support` package via Spago
- Updates pre-0.12 referent to `Control.Monad.Eff.Console` with `Effect.Console`.
- Update paths passed to `purs` in the section about launching a REPL without Spago.

### DIFF
--- a/guides/PSCi.md
+++ b/guides/PSCi.md
@@ -11,8 +11,6 @@ $ cd my-project              # Enter an empty folder
 
 $ spago init                 # Initialize a Spago environment
 
-$ spago install psci-support # Install PSCi support
-
 $ spago repl                 # Fire up the interpreter psci
 
 ...
--- a/guides/PSCi.md
+++ b/guides/PSCi.md
@@ -2,19 +2,24 @@ PSCi "PureScript Interactive" is the REPL for PureScript. It is often a good way
 
 ## Getting Started
 
-Use Pulp to configure and start PSCi:
+Use Spago to configure and start PSCi:
 
 ```text
-$ npm install -g pulp # Install pulp
+$ npm install -g spago       # Install Spago
 
-$ cd my-project # Enter an empty folder
+$ cd my-project              # Enter an empty folder
 
-$ pulp init # Initialize a pulp environment
+$ spago init                 # Initialize a Spago environment
 
-$ pulp repl # Fire up the interpreter psci
+$ spago install psci-support # Install PSCi support
 
-PSCi, version 0.9.1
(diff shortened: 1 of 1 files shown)


guides/PSCi.md
```diff
@@ -85,10 +90,10 @@ The PureScript compiler suite (i.e. the executable `purs`), unlike most compiler
 
 `purescript-psci-support` defines the `Eval` type class for this purpose. Instances of `Eval` are provided for `Show`able types, and for `Eff`, so that we can evaluate actions in the REPL. Library implementors might like to provide `Eval` instances for their own `Eff`-like types.
 
-## PSCi Without Pulp
+## PSCi Without Spago
 
 PSCi can be run directly, by specifying a list of PureScript source files as globs:
 
-    psci 'src/**/*.purs' 'bower_components/purescript-*/src/**/*.purs'
+    psci 'src/**/*.purs' 'output/**/*.purs'
```
natefaubion: The command now is `purs repl`. I think we should just omit the secondary lib glob since `output` isn't generally a lib glob; it's what the compiler uses for it's build products and won't work with `purs repl`.

hdgarrood:
Is the psci-support step necessary? I thought that Spago includes it by default now. Are you using an up to date version of Spago?

hdgarrood:
Yeah this was resolved 11 months ago, see https://github.com/purescript/spago/pull/174 , so I think I’d rather omit the`spago install  psci-support` step.

guides/PSCi.md
hdgarrood: Omitting the second one is still kind of problematic because it’s inevitably going to lead to a slew of module not found errors in almost any case it’s used (because the dependencies won’t be included). I’m not really sure what to do here; the way spago works means that there isn’t a single glob you can use here, you have to use `spago sources`, and I’m not sure it counts as “using the repl on its own” if you are still using Spago to assemble the command line arguments.

mbbx6spp:
@hdgarrood I remove the `spago install psci-support`. See fb63e01.

guides/PSCi.md
mbbx6spp: I didn't notice the `psci` command being there, apologies. I attempted to make a suggested change in 807183b to address these concerns plus updated more references to the old `psci` command to either use `spago repl` or `purs repl` which I missed before. Thanks.

hdgarrood:
Looks good, thank you!

PR #296, opened 2020-02-29
Update 'Editor and tool support' Guide to reflect current status

mbbx6spp:
This updates the list of build tools and package managers for PureScript in the 'Editor and tool support' guide based on last activity witnessed in linked repositories, splitting the tools into the categories:
 - maintained and actively supported
 - deprecated (due to repositories slightly staling)
 - unmaintained (due to either maintainer stating it was unmaintained or no commits on mainline in last 1.5 years)

### DIFF
--- a/ecosystem/Editor-and-tool-support.md
+++ b/ecosystem/Editor-and-tool-support.md
@@ -37,12 +37,17 @@ Spacemacs users can just use the [`purescript` layer](https://github.com/syl20bn
 
 ## Build tools and package managers
 
-- [Pulp](https://github.com/purescript-contrib/pulp) - a standalone build system for PureScript ([pulp](https://www.npmjs.com/package/pulp) in `npm`)
+Maintained and actively supported tools:
+- [spago](https://github.com/purescript/spago) - PureScript package manager and build tool powered by [Dhall](https://github.com/dhall-lang/dhall-lang) and [package-sets](https://github.com/purescript/package-sets)
 - [psc-package](https://github.com/purescript/psc-package) - A package manager for PureScript based on the concept of package sets
-- [spago](https://github.com/spacchetti/spago) - PureScript package manager and build tool powered by [Dhall](https://github.com/dhall-lang/dhall-lang) and [package-sets](https://github.com/purescript/package-sets)
-- [Gulp task](https://github.com/purescript-contrib/gulp-purescript) (`gulp-purescript` in `npm`)
-- [psvm-js](https://github.com/ThomasCrvsr/psvm-js) - PureScript Version Manager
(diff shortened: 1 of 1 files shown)


ecosystem/Editor-and-tool-support.md
```diff
@@ -37,12 +37,17 @@ Spacemacs users can just use the [`purescript` layer](https://github.com/syl20bn
...
+- [pscid](https://github.com/kRITZCREEK/pscid) is a lightweight file-watcher/testrunner for PS projects providing instant-rebuilds in an editor agnostic way
+
+Deprecated:
+- [Pulp](https://github.com/purescript-contrib/pulp) - a standalone build system for PureScript ([pulp](https://www.npmjs.com/package/pulp) in `npm`). *Spago is recommended over Pulp.*
+- [Gulp task](https://github.com/purescript-contrib/gulp-purescript) (`gulp-purescript` in `npm`). *Spago is recommended over the PureScript Gulp task.*
+
+Unmaintained:
+- [psvm-js](https://github.com/ThomasCrvsr/psvm-js) - PureScript Version Manager
 - [psc-pane](https://github.com/anttih/psc-pane) - Auto reloading compiler which formats a single error to fit the window
 - [purescript-psa](https://github.com/natefaubion/purescript-psa) - A pretty, flexible error/warning reporting frontend for `psc` featuring colours, original source spans in errors, warning filtering and persistence.
```
natefaubion: `psa` is not unmaintained, it just does everything it needs to and does it well. It's the first thing we recommend when users ask for better error output and configuration of warnings.

natefaubion: Additionally, `psa` is backed by `psa-utils` which `pscid` and the lsp implementations rely on.

hdgarrood:
Thanks for this - this page definitely does need updating. However, I don’t think it’s safe to designate things as ‘unmaintained’ or even ‘deprecated’ just on the basis of GitHub history. Along the same lines - my understanding of the word “deprecation” is that it means that the authors or maintainers have stated that they do not intend to continue to work on the thing and that they recommend using an alternative, so it isn’t a label that you can really apply to other people’s projects without their input.

Pulp isn’t deprecated either. I do plan to deprecate it eventually but at the moment it is the only tool I’m aware of which can publish packages to Pursuit, and we can’t deprecate it until there’s an alternative (which will likely be Spago, soon).

I’ll have a think about how I’d like to proceed here and get back to you.

mbbx6spp:
@hdgarrood @natefaubion Totally fair. I added a new change and kept only two in the unmaintained list (one is stated as not maintained by the author and the other has not had any activity in 2.5 years with outstanding issues for 4 years also). I moved the `psa` link up to maintained. See 7d291e9. Happy to make more changes too.

PR #298, opened 2020-02-29
Replace references to pulp with spago in error docs

mbbx6spp:
Updates error docs to use `spago` instead of `pulp` for consistency across all docs.


hdgarrood:
Thanks!

PR #299, opened 2020-02-29
Update 'Getting Started' guide to include new installation methods

mbbx6spp:
Updates 'Getting Started' guide with different subsections for existing and new methods of installation, including:
 - Using NPM
 - Downloading Binaries
 - Homebrew
 - Nix

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -4,7 +4,9 @@ Let's walk through the basics of getting set up to use the PureScript compiler `
 
 We'll start with the installation of the compiler and Spago build tool, and then go through the basic usage of `purs repl`, working towards a solution of problem 1 from [Project Euler](http://projecteuler.net/problem=1).
 
-#### Installing the Compiler
+### Installing the Compiler
+
+#### Using NPM
 
 You'll need [Node.js and npm](https://docs.npmjs.com/getting-started/installing-node) and to be [able to install global packages](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-1-change-the-permission-to-npm-s-default-directory) to proceed.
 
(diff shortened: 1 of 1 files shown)


mbbx6spp:
Related to purescript/purescript.github.io#141 which I will be modifying momentarily to account for this change.

hdgarrood:
I’m not sure about this one; personally, I’m most concerned about the users who don’t know or care what all of these different install options are, and I definitely don’t want to overload them with information they don’t know what to do with and present them with choices they aren’t equipped to make. Of course this information should be recorded somewhere, but the purpose of this guide is to help people get a basic setup working with absolutely minimal fuss, so I am not sure here is the correct place. There’s INSTALL.md in the compiler right now, which covers alternative installation methods to some extent, but that is admittedly very easy to overlook.

mbbx6spp:
@hdgarrood would it make sense for me to relocate these new install methods in the INSTALL.md of the compiler and link to there from this document for those interested?

hdgarrood:
Yes, that sounds good to me.

guides/Getting-Started.md
```diff
@@ -4,17 +4,31 @@ Let's walk through the basics of getting set up to use the PureScript compiler `
...
-#### Setting up the Development Environment
+#### Using Homebrew
+
+    brew install purescript
+
+#### Using Nix
+
+You'll need [Nix](https://nixos.org/nix/) installed with a recent version of `nixpkgs` (19.09):
+
+    nix-env -i purescript
```
toraritte, line 29: May I suggest also adding [`easy-purescript-nix`](https://github.com/justinwoo/easy-purescript-nix)? Just starting learning PureScript, and it has everything a beginner needs (or any PS dev for that matter) - plus nothing is broken as it pins nixpkgs. [This stackoverflow thread](https://stackoverflow.com/questions/46094241/need-help-on-start-using-purescript-on-nixos) is a perfect summary how confusing the PS eco-system for a new-comer, adding NixOS to mix as well.

toraritte: It may be worth mentioning it not just in [INSTALL.md](https://github.com/purescript/purescript/blob/master/INSTALL.md) but somewhere here in this documentation repo, because `easy-purescript-repo` sets up a(n almost) complete development environment with the ["recommended" utilities](https://discourse.purescript.org/t/recommended-tooling-for-purescript-in-2020/1615).

thomashoneyman:
This also relates to #323

hdgarrood:
We're going to use INSTALL.md in the compiler repo to discuss other installation methods, so I'll close this. As far as I can tell, the compiler's INSTALL.md covers all of the mentioned resources apart from Nix (which I am open to taking a PR for).

PR #301, opened 2020-03-05
purerl/purescript repo is archived

zoechi:
with no obvious link to the new repo


hdgarrood:
Thanks! I think in that case we might as well just update the main link to point straight to `purerl/purerl` then.

hdgarrood:
I think this was addressed by #307.

PR #303, opened 2020-03-26
Update Records.md

flip111:
I just get started with purescript and the documentation is too minimal for me and doesn't give the bridge to the rest of the ecosystem. I feel like adding two more examples with increasing complexity fills a void.

### DIFF
--- a/language/Records.md
+++ b/language/Records.md
@@ -76,8 +76,31 @@ setPersonPostcode pc p = p { address { postCode = pc } }
 A record update function can also be defined by using an `_` inplace of the record to be updated like:
 
 ```purescript
-_ { fieldName = newValue }
+inPlace :: Value -> Record -> Record
+inPlace newValue = _ { fieldName = newValue }
 ```
+
+A record defined with `data` first has to be unwrapped
+
+```purescript
+data PointRec = PointRec { x :: Number, y :: Number }
+
+setX :: Number -> PointRec -> PointRec 
+setX val (PointRec point) = PointRec (point { x = val })
+```
+
+A more complex data structure can be updated by using intermediate getters. `r.fieldTwo` in the example below
+
+```purescript
+type Example = { fieldOne :: Boolean, fieldTwo :: Maybe { innerField :: Int} }
+
+addToInner :: Int -> Example -> Example
+addToInner n r = case r.fieldTwo of
+                   Nothing -> r
+                   Just k  -> r { fieldTwo = Just (k.innerField + n) }
+```
+
+For more complex record updates a package like [profunctor-lenses](https://github.com/purescript-contrib/purescript-profunctor-lenses) can be [useful](https://thomashoneyman.com/articles/practical-profunctor-lenses-optics/).
 
 ## Field Names
 


hdgarrood:
Thanks for the PR, but this is not really in line with the aims of this repo. These docs are more reference documentation, they're not intended to function as a tutorial -- that's what the [book](https://github.com/purescript-contrib/purescript-book) is for.

flip111:
The book does not cover record update very well (it only mentions this in two places). At the moment there are gaps between the purescript documentation and the book. This seems to be the easiest place to fill some of the gap because the documentation has a chapter for it.

The book could mimic the structure of the language reference (only the parts for which additional content is written) and then put this in an appendix or so.

Can the language ref and the book then link to each other on the respective chapters?

In any case it seems that some more discussion is needed first, which is a unfortunate situation because i can't make the contribution now.

JordanMartinez:
@flip111 I have something written here: https://github.com/JordanMartinez/purescript-jordans-reference/blob/latestRelease/11-Syntax/01-Basic-Syntax/src/02-Data-and-Functions/09-Records/01-Basic-Syntax.purs#L93-L123

Also, I have something written here that explains that one should not be using `data` when defining a record: https://github.com/JordanMartinez/purescript-jordans-reference/blob/latestRelease/31-Design-Patterns/04-Records--Use-Type-Data-or-Newtype.md

If you want to add your changes to my repo's syntax folder, I'd be fine with that.

flip111:
Hi @JordanMartinez that's nice :) 

It's good that there are multiple resources. Seems that yours is not referenced by the documentation or the purescript book. I like all the contributions people have many in various places, they just seems to be a bit scattered / unlinked at the moment.

hdgarrood:
As it happens Jordan's repo is referenced by this one, it's in the README.

flip111:
I meant fine grained linking on the subject of record updates or other subjects of similar granularity.

JordanMartinez:
> I meant fine grained linking on the subject of record updates or other subjects of similar granularity.

Mm... the content in the Syntax folder is not as fine-grained as it could be. That file might actually have that, but most don't :sweat_smile: It's not hard to fix that. I just don't want to overload my project's ToC with so many entries that it's useless.

> As it happens Jordan's repo is referenced by this one, it's in the README.

And [it's not too far into the ReadMe either](https://github.com/purescript/documentation#learning). @flip111 Is this the first time you've heard of my repo?

flip111:
> @flip111 Is this the first time you've heard of my repo?

Yes, i use google all the time to find stuff about purescript and yours never comes to the top results.

JordanMartinez:
@flip111 JordanMartinez/purescript-jordans-reference#338

Perhaps more would come up if I converted the repo into a state site via GH Pages...

flip111:
@JordanMartinez [Here](https://discourse.purescript.org/t/how-to-create-a-javascript-object-with-a-value-that-doesnt-have-a-fixed-type/1233/4?u=flip101) @hdgarrood gives an advice `You can always use forall props. Record props, which means “a record with any fields which may have any type”`. After reading your guide [here](https://github.com/JordanMartinez/purescript-jordans-reference/blob/latestRelease/11-Syntax/01-Basic-Syntax/src/02-Data-and-Functions/09-Records/03-Row-Polymorphism.purs) i still don't get how this can work. Maybe we can make some addition to the guide?

JordanMartinez:
@flip111 Let's move this discussion to my repo. I'd like to stop (as others, I'm sure, see it) spamming this PR with things that no longer directly relate to this repo.

PR #305, opened 2020-04-10
Add type declaration to test file to supress warnings

danny-andrews:
On the current version of purescript, omitting a type declaration for the
top-level declaration of main emits an MissingTypeDeclaration error. This
prevents that.

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -202,9 +202,11 @@ module Test.Main where
 
 import Prelude
 
+import Effect (Effect)
 import Euler (answer)
 import Test.Assert (assert)
 
+main :: Effect Unit
 main = do
   assert (answer == 233168)
 ```


hdgarrood:
From about two paragraphs before this:

> The compiler will display several warnings about missing type declarations. In general it is considered good practice to provide explicit type signatures. In this guide, they are left out for brevity. In the absence of type signatures, the PureScript compiler infers types automatically but will remind us to consider adding them.

hdgarrood:
To clarify my position, I'm open to adding type signatures so that these warnings aren't emitted, but also I'm tempted to leave it as it is. In any case, we should ensure that the guide is internally consistent, both in that we are using the same style throughout and in that it doesn't contradict itself by saying it won't use type signatures and then using them anyway.

danny-andrews:
I gotcha. As a complete beginner, seeing a warning in the console when I run the tests makes me feel like I did something wrong, but if I had seen the note, I would have been able to disregard it. So feel free to close this.

hdgarrood:
In practice almost all code does include type signatures, and I agree that having the expected path through this guide emit warnings is not ideal, so I think adding them in this guide is probably not a bad option. I think they might have been omitted originally because the old `Eff` type (which had two type arguments: an effect row, and the type of the result) could have been a bit intimidating, but now we just have `Effect` instead I think it's less of a concern. I would definitely be happy to consider adding type signatures throughout this guide (and removing that note that says they are omitted).

hdgarrood:
I'm going to close this for now as a bit more work would be necessary before this is mergeable, but we can continue discussion in #318.

PR #308, opened 2020-04-19
Remove reference to Eval instances for psci.

natefaubion:
It's not possible to write `Eval` instances due to the catch-all instance chain.


natefaubion:
Hah, of course. Done.

PR #313, opened 2020-05-08
Explain why Orphan Instances are forbidden

milesfrain:


language/Type-Classes.md
```diff
@@ -78,7 +78,7 @@ assert false = fail "Assertion failed"
 
 ## Orphan Instances
 
-Type class instances which are defined outside of both the module which defined the class and the module which defined the type are called *orphan instances*. Some programming languages (including Haskell) allow orphan instances with a warning, but in PureScript, they are forbidden. Any attempt to define an orphan instance in PureScript will mean that your program does not pass type checking.
+Type class instances which are defined outside of both the module which defined the class and the module which defined the type are called *orphan instances*. Only one instance is allowed per type and class combination. Orphan instances enable duplicate instances to be defined in separate modules. These modules are fine independently, but if both modules are ever imported into the same project, then an instance collision would break the build. Some programming languages (including Haskell) allow orphan instances with a warning, but in PureScript, they are forbidden. Any attempt to define an orphan instance in PureScript will mean that your program does not pass type checking.
```
milesfrain, line 81: Added this section:
> Only one instance is allowed per type and class combination. Orphan instances enable duplicate instances to be defined in separate modules. These modules are fine independently, but if both modules are ever imported into the same project, then an instance collision would break the build.

The diff makes this difficult to see.

hdgarrood:
I think the explanation of why they are forbidden should come after the first paragraph, rather than occur within it; we need to wait until the reader knows what an orphan instance is and what will happen if they try to define one before we try to explain why this happens.

Also, there's more to it than just the issue that the compiler won't know what to do if two different choices for a particular instance are in scope. For example, banning orphan instances gives us global uniqueness of instances: the property that there can only ever exist one instance for a particular class and type (or set of types, for multi-parameter classes). This property is useful in quite a few places; you may have made use of it unknowingly. One example is of Ord-based maps and sets: if it were possible to insert some values into a map using one Ord instance, and then try to retrieve them using a different one, you'd have keys just disappear from your map. Another example is if you had a type class which defined serialization and deserialization operations: global uniqueness of instances means that you can't serialize something with one instance and then try to deserialize it with a different one.

milesfrain:
The line of reasoning that makes most sense for me is:
1. What can happen without any rules (overlapping imports)
1. What contributes to this overlapping issue (orphan instances)
1. What the compiler does when it detects orphan instances (error)

But I adjusted this PR based on your feedback.

language/Type-Classes.md
```diff
@@ -106,6 +106,10 @@ instance semigroupAddInt :: Semigroup AddInt where
 
 In fact, a type similar to this `AddInt` is provided in `Data.Monoid.Additive`, in the `monoid` package.
 
+The reason why orphan instances are banned is because they can lead to two types of problems:
+* Recall that only one instance is allowed per type and class combination in a module. Orphan instances enable duplicate instances to be defined in separate modules without breaking this overlapping instances rule. So these modules are fine independently, but if both modules are ever imported into the same project (for example from separate libraries), then an instance collision would break the build.
```
hdgarrood, line 110: This bullet point moves back and forth between the way things actually are (i.e. orphans are banned) and the way things might be if they weren't, and so I think it could be made a bit tighter by addressing that. Additionally, I think a slightly more explicit example would help. How about:

> Suppose two separate modules define an orphan `Semigroup Int` instance, and one of them uses `+` for `append`, whereas the other uses `*`. Now suppose someone writes a third module which imports both of the first two, and that somewhere in that third module we have the expression `2 <> 3`, which calls for a `Semigroup Int` instance. The compiler now has two instances to choose from. What should it do? It could report an error, or it could arbitrarily pick one of the instances. Neither option is particularly appealing:
>
> * If it chooses to report an error, it means that any pair of modules which define the same orphan instance can never be used together.
> * If it arbitrarily picks one, we won't be able to determine whether `2 <> 3` will evaluate to 5 or 6. This can make it very difficult to ensure that your program will behave correctly!

milesfrain: That example ties in nicely with the previous code snippet. Included it in the latest push.

hdgarrood: Thanks! That's why I used it :slightly_smiling_face:

language/Type-Classes.md
```diff
@@ -106,6 +106,13 @@ instance semigroupAddInt :: Semigroup AddInt where
...
+
+ * If it chooses to report an error, it means that any pair of modules which define the same orphan instance can never be used together.
+ * If it arbitrarily picks one, we won't be able to determine whether `2 <> 3` will evaluate to 5 or 6. This can make it very difficult to ensure that your program will behave correctly!
+
+Banning orphan instances also ensures global uniques of instances. Without global uniques, you risk operating on data with incompatible instances in different sections of code. For example, in Ord-based maps and sets, if it were possible to insert some values into a map using one `Ord` instance, and then try to retrieve them using a different `Ord` instance, you'd have keys disappear from your map. Another example is if you had a type class which defined serialization and deserialization operations, you'd be able to serialize something with one instance and then try to deserialize it with a different incompatible instance.
```
hdgarrood, line 114: Oh sorry, one last thing - I just noticed "global uniques" versus "global uniqueness". Was that a deliberate change? "global uniqueness" makes more sense to me, personally.

hdgarrood:
Thanks - there's one more occurrence of "global uniques" in the following sentence, too.

Issue #309, opened 2020-04-21
Merge getting-started guide with purescript book


milesfrain:
How do we feel about merging the [getting-started guide in this repo](https://github.com/purescript/documentation/blob/master/guides/Getting-Started.md) with the [getting started chapter in the book](https://book.purescript.org/chapter2.html)?

I'm hoping that deduplicating the guides will make maintenance easier and be better for new users.

If all content of this repo's guide is incorporated into the book, could we just redirect the "Quick Start Guide" link on [purescript.org](https://www.purescript.org/) to the book's chapter?

Wondering which example should be used. The Euler problem is more interesting, but the books diagonal function is simpler for beginners. Could also keep both examples.

oldfartdeveloper:
I agree that the important task is to have only one "getting-started-guide" w/ no duplication in multiple places.

I'm slightly inclined to move the getting-started-guide into the purescript book and redirect the "Quick Start Guide".

In the getting-started-guide, I prefer using the "diagonal function" for the reason that I'm just trying to get the environment up and running; I don't care about interesting programs at that particular time.  The explanation of the generated Javascript is more cumbersome for the Euler exercise and I don't think the extra effort is appropriate in that place.

But since the Euler problem IS interesting, I'd like to see it drawn in to illustrate some aspect of Purescript that the book explains after the getting-started-guide.

Since I've already made some suggestions for the getting-started-guide (#324), I'm in this general area anyways and would be glad to follow up here by implementing this issue's objectives.

milesfrain:
I agree with all your points.

Reasons for going with the current arrangement can be found in 
https://github.com/purescript-contrib/purescript-book/issues/93

We could always reconsider this decision though, although it's not the highest priority for me now that we accomplished the deduplication goals with merges in both repos.

PR #314, opened 2020-05-16
Incorporate getting-started content from book

milesfrain:
There's an accompanying PR (https://github.com/purescript-contrib/purescript-book/pull/144) in the book repo to point to this guide as a first step in environment setup.

More discussion about this change can be read in: https://github.com/purescript-contrib/purescript-book/issues/93

This address #309 in this repo.


guides/Getting-Started.md
```diff
@@ -6,13 +6,17 @@ We'll start with the installation of the compiler and Spago build tool, and then
 
 #### Installing the Compiler
 
-You'll need [Node.js and npm](https://docs.npmjs.com/getting-started/installing-node) and to be [able to install global packages](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-1-change-the-permission-to-npm-s-default-directory) to proceed.
+You'll need to install [Node.js and npm](https://docs.npmjs.com/getting-started/installing-node).  We recommend installing [Node.js and npm via a "node version manager"](https://docs.npmjs.com/getting-started/installing-node) to avoid issues with installing packages gloablly. If you choose to install it manually, you might experience the [`EACCES` error when installing packages globally](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-1-change-the-permission-to-npm-s-default-directory).
```
hdgarrood, line 9: Following npm's recommendation here sounds good to me, but a style nit: could you please remove the quotes from "node version manager"?

guides/Getting-Started.md
```diff
@@ -6,13 +6,17 @@ We'll start with the installation of the compiler and Spago build tool, and then
...
 
-The Purescript compiler (`purs`) can be installed with npm:
+Install the Purescript compiler (`purs`) with npm:
```
hdgarrood, line 11: There have been a few cases where leaving it implicit that the guide actually expects you to do things (versus just letting you know that they are possible) has caused confusion, so :+1: for explicitly telling the reader to do things like this.

guides/Getting-Started.md
```diff
@@ -6,13 +6,17 @@ We'll start with the installation of the compiler and Spago build tool, and then
...
 
     npm install -g purescript
 
-(It can also be installed from [Hackage](http://hackage.haskell.org/package/purescript), or by downloading the latest [binary bundle](https://github.com/purescript/purescript/releases) for your OS. If you do so, make sure the `purs` executable is on your `$PATH`.)
+Try running the PureScript compiler on the command line to verify that the PureScript compiler executables are available on your path:
```
hdgarrood, line 15: This sentence uses "path", but the following one uses `$PATH`. I think we should just pick one, so that it's clear we're referring to the same thing in each case. I'd recommend `$PATH`, as it ought to be clearer we're using the term in a specific technical sense that way.

guides/Getting-Started.md
```diff
@@ -231,10 +244,84 @@ main = do
...
+#### Compiling for the Browser
+
+Spago can be used to turn our PureScript code into JavaScript suitable for use in the web browser by using the `spago bundle-app` command:
+
+    $ spago bundle-app
+    ...
+    Build succeeded.
+    Bundle succeeded and output file to index.js
+
+All the code in the `src` directory, a standard PureScript library known as the _Prelude_ and any project dependencies have been compiled to JavaScript. The resulting code is bundled as `index.js` and has also had any unused code removed, a process known as dead-code-elimination. This `index.js` file can now be included in an HTML document. If you try this, you should see the words "Hello, World!" printed to your browser's console.
```
hdgarrood, line 261: We've already introduced the Prelude by this point, so I think the phrase "a standard PureScript library known as the Prelude" is unnecessary here.

hdgarrood: Also, while I realise it was like this in the book to start with, would you mind replacing the hyphens in "dead-code-elimination" with spaces? I don't think hyphens make sense here grammatically.

guides/Getting-Started.md
hdgarrood: Also, there's a typo here: "gloablly".

milesfrain:
Thanks for the thorough review. Applied feedback in latest push.

hdgarrood:
Also, it's not a problem now, but would you mind pushing new commits rather than rebasing and force-pushing in the future? It's easier to review what's changed since the last time I looked that way, and GitHub already makes it easy to squash when it comes to merging.

milesfrain:
Yes, I agree. I believe this one was an incremental commit though.

hdgarrood:
:facepalm: so it was, I just don't know how to use the GitHub UI apparently. Sorry!

Issue #315, opened 2020-05-16
Point "PuresScript by Example" links to book.purescript.org


kurtmilam:
By my count, there are are five links in the documentation that point to the book on leanpub.

I'd be happy to open a PR if there's interest.

Pages with links to leanpub (may not be exhaustive):
- [README.md#project-scope](https://github.com/purescript/documentation#project-scope)
- [guides/FFI-Tips.md#why-doesnt-my-eff-work-when-passed-to-a-normal-js-function](https://github.com/purescript/documentation/blob/master/guides/FFI-Tips.md#why-doesnt-my-eff-work-when-passed-to-a-normal-js-function)
- [guides/FFI.md#conclusion](https://github.com/purescript/documentation/blob/master/guides/FFI.md#conclusion)
- [language/Type-Classes.md#multi-parameter-type-classes](https://github.com/purescript/documentation/blob/master/language/Type-Classes.md#multi-parameter-type-classes)
- [language/Type-Classes.md#functional-dependencies](https://github.com/purescript/documentation/blob/master/language/Type-Classes.md#functional-dependencies)

hdgarrood:
Yes please!

thomashoneyman:
This can be closed now that #316 has been merged.

PR #316, opened 2020-05-16
Point Book Links to book.purescript.org

kurtmilam:
This PR updates five links to "PureScript by Example", pointing them at [book.purescript.org](https://book.purescript.org) rather than leanpub.

Related issue: https://github.com/purescript/documentation/issues/315

### DIFF
--- a/guides/FFI-Tips.md
+++ b/guides/FFI-Tips.md
@@ -117,7 +117,7 @@ By moving the `show` reference out to `showSomething` the compiler will pick the
 
 ## Why Doesn't my `Eff` Work When Passed to a Normal JS Function?
 
-["Representing Side Effects"](https://book.purescript.org/chapter8.html#the-effect-monad-1) in *PureScript by Example*.
+["Representing Side Effects"](https://book.purescript.org/chapter10.html#representing-side-effects) in *PureScript by Example*.
 
 In order to avoid prematurely evaluating effects (or evaluating effects that should not be evaluated at all), PureScript wraps them in constant functions:
 
--- a/guides/FFI-Tips.md
+++ b/guides/FFI-Tips.md
@@ -117,7 +117,7 @@ By moving the `show` reference out to `showSomething` the compiler will pick the
 
 ## Why Doesn't my `Eff` Work When Passed to a Normal JS Function?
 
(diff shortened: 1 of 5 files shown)


guides/FFI-Tips.md
```diff
@@ -117,7 +117,7 @@ By moving the `show` reference out to `showSomething` the compiler will pick the
 
 ## Why Doesn't my `Eff` Work When Passed to a Normal JS Function?
 
-["Representing Side Effects"](https://leanpub.com/purescript/read#leanpub-auto-representing-side-effects) in *PureScript by Example*.
+["Representing Side Effects"](https://book.purescript.org/chapter8.html#the-effect-monad-1) in *PureScript by Example*.
```
kurtmilam, line 120: I wasn't 100% sure to which section in this chapter I should link. This was my best guess.

kurtmilam:
I could also update `Eff` to Effect` in[ this section of the FFI Tips](https://github.com/purescript/documentation/blob/master/guides/FFI-Tips.md#why-doesnt-my-eff-work-when-passed-to-a-normal-js-function), but that may also require some more thought:


## Why Doesn't my `Eff` Work When Passed to a Normal JS Function?

["Representing Side Effects"](https://book.purescript.org/chapter8.html#the-effect-monad-1) in *PureScript by Example*.

In order to avoid prematurely evaluating effects (or evaluating effects that should not be evaluated at all), PureScript wraps them in constant functions:

```javascript
exports.myEff = function() {
  return doSomethingEffectful(1, 2, 3);
}
```

which is imported to PureScript as:

```purescript
foreign import myEff :: forall eff. Eff (myEff :: MYEFF | eff) SomeType
```

guides/FFI-Tips.md
hdgarrood: I think https://book.purescript.org/chapter10.html#representing-side-effects would be more appropriate, as that's the one which deals with the runtime representation of `Effect`.

hdgarrood:
I think it would probably be best to not worry about updating Eff to Effect in this PR.

guides/FFI-Tips.md
kurtmilam: done!

hdgarrood:
Thanks!

PR #317, opened 2020-05-17
Guard Limitations in `let`

milesfrain:


hdgarrood:
From what I can tell from looking at the parser, it’s not just constructor patterns that don’t work with guards, it’s any pattern other than a simple identifier. For example, it looks to me like record and array patterns and at-patterns (`foo@(...)`) don’t work in this context either. I also think we ought to be able to come up with a better justification for this. I’ll have a think.

natefaubion:
It's that way "because". See https://github.com/purescript/purescript/pull/3200

milesfrain:
Should we add the caveat, then open a new issue to remove it?

language/Pattern-Matching.md
```diff
@@ -176,6 +176,36 @@ compare _ _ = EQ
...
+-- This doesn't work
+f1 :: Int
+f1 =
+  let
+    (Tuple a b)
+      | false = Tuple 1 2
+      | otherwise = Tuple 3 4
+  in
+    a
+```
```
thomashoneyman, line 193: Some of the capitalization is a bit odd -- we don't usually capitalize "guard" or "constructor". What about:

> Constructor patterns with guards are not supported in `let` expressions. For example, this does not compile:

We could also trim this example a little -- the example from https://github.com/purescript/purescript/pull/3200 is pretty good:

```purs
f :: Int -> Int -> Int
f a b =
  let 
    Tuple x y
      | a < b = Tuple a b
      | otherwise = Tuple b a
  
  in x
```

> A workaround is to separate the guard from the constructor pattern:

```purs
f :: Int -> Int -> Int
f a b =
  let 
    tup
      | a < b = Tuple a b
      | otherwise = Tuple b a

    Tuple x y = tup
  
  in x
```

thomashoneyman: Other than this I think it's sensible to merge this while the restriction does exist.

language/Pattern-Matching.md
```diff
@@ -176,6 +176,36 @@ compare _ _ = EQ
...
+
+A workaround is to separate the constructor pattern from the Guard:
```
thomashoneyman, line 195:
```suggestion
A workaround is to separate the constructor pattern from the guard:
```

JordanMartinez:
This PR hasn't been merged because it's a bit incomplete, I believe.

PR #321, opened 2020-05-23
Add note on Function naming

milesfrain:


language/Syntax.md
```diff
@@ -97,6 +97,13 @@ multiply x y = x * y
 
 Type signatures are not required for top-level declarations in general, but is good practice to do so. See the section on types for more details.
 
+## Function names
```
hdgarrood, line 100: We should call this section "value names", since these rules apply to all values, not just functions. Eg

```
-- not ok, starts with a number
10 = 10

-- ok
_10 = 10
```

language/Syntax.md
```diff
@@ -97,6 +97,13 @@ multiply x y = x * y
...
+
+Function names must start with a lowercase letter ([unicode category](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category) `Ll`) and may be followed by any number of:
```
hdgarrood, line 102: I've realised that I missed a couple of cases in my description before: in addition to what you've got here, value names may also start with an underscore, and may also be followed by underscore or apostrophe (U+0027) `'`.

hdgarrood:
Nice work figuring out the actual categories, by the way. How did you do it, just trial and error? The documentation for `Data.Char.isAlphaNum` and `Data.Char.isLower` leaves a bit to be desired: https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Char.html#v:isAlphaNum

language/Syntax.md
milesfrain: I thought values were just "functions of zero arguments".

hdgarrood: I mean, you can think of them like that if you want to, but it’s not that helpful a perspective in my view.

milesfrain:
> How did you do it?

I actually mistakenly looked it at the [PS `isAlphaNum`](https://pursuit.purescript.org/packages/purescript-unicode/4.0.1/docs/Data.Char.Unicode#v:isAlphaNum), then found the [categories listed here](https://github.com/purescript-contrib/purescript-unicode/blob/bb70f5a17aa9bcad05104c5a3f191fd75bb3d2ce/src/Data/Char/Unicode/Internal.purs#L4834-L4838). Hopefully this is similar to the HS version.

I'll make the edits you described once I hear back about "values" vs "functions of zero arguments".
I guess it's more accurate to say that functions only have one argument, and that functions are values with an arrow type.
But it seems like you can also describe functions of "zero or more arguments" - at least that's what I did at the end of [this post](https://discourse.purescript.org/t/purescript-book-questions/1334/3).

hdgarrood:
Yes - if we are being precise, all PureScript functions have exactly one argument; functions of “more than one argument“ are really just functions which return functions. However “function of N arguments” for N > 1 is almost universally understood and is a useful shorthand, so it is used too. “Function of 0 arguments” is a bit odd; it’s easier to just say “value”.

JordanMartinez:
For what it's worth "Functions with 0 arguments" on the JS side (e.g. `function () { return 0; }`) have the type, `Unit -> ReturnValue`.

language/Syntax.md
```diff
@@ -97,6 +97,15 @@ multiply x y = x * y
...
 Type signatures are not required for top-level declarations in general, but is good practice to do so. See the section on types for more details.
 
+## Function and Value names
+
+Function and value names must start with a lowercase letter ([unicode category](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category) `Ll`) or underscore `_`. The following characters may be any number of:
+* Letter: Unicode category `L`
+* Mark: Unicode category `M`
+* Number: Unicode category `N`
+* Underscore: `_`
+* Apostrophie: `'`
```
hdgarrood, line 107: Typo here: “Apostrophe”

PR #322, opened 2020-05-28
Add instructions for export default in purescript

realvictorprm:
Took me quite a while to figure this one out. Maybe that simple but useful bit is well placed here?

### DIFF
--- a/language/Modules.md
+++ b/language/Modules.md
@@ -183,3 +183,15 @@ module Test (class Foldable, foldr, foldl, foldMap) where
 ```
 
 If a type class is exported, then all of its members must also be exported. Likewise, if a type class member is exported, the type class it belongs to must also be exported.
+
+### Default export a function
+
+To mimic the `export default` for a binding add a binding `default` and add it to the export list of the module:
+
+```purescript
+module Component(default) where
+
+someComponent = ...
+
+default = someComponent
+```


hdgarrood:
I'm not sure about this: I'm not sure which `export default` you're referring to here, and I also think that this functionality is covered by the existing "Module exports" subheading. I understand that it can be frustrating that things are under-documented but this repository is only meant as reference-style documentation for people who already have some familiarity with the language (realistically, we don't have the means to maintain more beginner-style docs here as well).

realvictorprm:
Hm ok. If this is the wrong place, do you have recommendations where this could be put?

hdgarrood:
Maybe https://github.com/JordanMartinez/purescript-jordans-reference ? I think part of what makes it hard to decide where to put it is that this is more of a stylistic preference, and I don't think this style is particularly idiomatic in PureScript. Personally I'd just export `someComponent` under that name.

JordanMartinez:
Yeah, why not just export `someComponent`?

realvictorprm:
Hey folks :)

Just for clarification. In my specific use case it took me some time to figure out how to do the equivalent of `export default ...` from ES6 in Purescript because a scenario I was using required that the export is a `default` export. Otherwise I would have loved to use a different export name too :(

realvictorprm:
Side note. It might be obvious to people which have a lot of experience with web development. However I have a background in backend development and "classic" application development and so the export stuff and differences styles of `export` (ES5, ES6) and it's meaning weren't completely clear to me from beginning :/

hdgarrood:
Ahhh, I see. So I was wrong, it's not really a stylistic preference.

FWIW I think the fact that `export default` is treated the same as a named export called "default" is a quirk of how current bundlers are implementing ESM, and isn't something that would be true with a real ESM implementation, so it wouldn't be safe to rely on in general.

realvictorprm:
Lucky me for finding this out. Otherwise the stuff I used today wouldn't have worked out 😅 .

Should I rephrase it in this case with an addition that it's `unstable` and may change in future?

hdgarrood:
Actually, I think I was mistaken, I think `export default` is the same as a named export called "default" after all. See https://tc39.es/ecma262/#sec-static-semantics-importentriesformodule

rainbyte:
> Hey folks :)
> 
> Just for clarification. In my specific use case it took me some time to figure out how to do the equivalent of `export default ...` from ES6 in Purescript because a scenario I was using required that the export is a `default` export. Otherwise I would have loved to use a different export name too :(

Thanks for posting this information. I had the same problem before, and couldn't understand why.

realvictorprm:
Reassuring to hear that I'm not the only one experiencing this😅

realvictorprm:
@hdgarrood so should I make some changes or do you think this is worth to merge?

hdgarrood:
I think this is too JS-specific to go in the language reference, so it would be better suited to a JS interop guide.

hdgarrood:
I'm going to close this for now, but thanks anyway for the PR.

PR #323, opened 2020-06-09
Point to compiler's install guide

milesfrain:
Boosting heading sizes too so they're slightly larger than bold text.

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -4,7 +4,7 @@ Let's walk through the basics of getting set up to use the PureScript compiler `
 
 We'll start with the installation of the compiler and Spago build tool, and then go through the basic usage of `purs repl`, working towards a solution of problem 1 from [Project Euler](http://projecteuler.net/problem=1).
 
-#### Installing the Compiler
+### Installing the Compiler
 
 You'll need to install [Node.js and npm](https://docs.npmjs.com/getting-started/installing-node).  We recommend installing [Node.js and npm via a node version manager](https://docs.npmjs.com/getting-started/installing-node) to avoid issues with installing packages globally. If you choose to install it manually, you might experience the [`EACCES` error when installing packages globally](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-1-change-the-permission-to-npm-s-default-directory).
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
Thanks, but I don't think this should be part of the main getting started guide. I'd prefer a link to the INSTALL.md file in the compiler repo, which covers installation FAQs.

milesfrain:
> which covers installation FAQs.

Does this FAQ need to be created?
https://github.com/purescript/purescript/blob/master/INSTALL.md

The motivation for a troubleshooting section here is that we often need to help beginners through EACCES and libtinfo errors on slack. For example, search:
```
in:#purescript-beginners in:#purescript EACCES
in:#purescript-beginners in:#purescript libtinfo
```

Given that lots of users already miss the EACCES note in these instructions, I'm not sure how likely they'll be to find their solution in a linked FAQ. Hoping these prominent error snippets on this same page will help them see the connection.

hdgarrood:
I don’t understand your question about whether it needs to be created? That INSTALL.md file you linked already covers some common questions and issues that arise for people trying to install the compiler, such as supported OS versions, how to build it from source, and installing ncurses.

I don’t think we can assume that people are reading this guide and missing the existing stuff. There are many “getting started” resources out there aside from this one, or people might not even be using any, they might just be running `npm install -g purescript` (as suggested on purescript.org, for instance). I think a link to the INSTALL.md file can (and should) be prominent here without having to take up space in this file describing specific problems and solutions.

hdgarrood:
I think the most effective approach would probably be to get the npm installer to try to identify these issues where they occur and print appropriate error messages.

milesfrain:
> I don’t understand your question about whether it needs to be created?

I meant whether a specific "FAQ" or "Troubleshooting" section should be created in INSTALL.md to help users more easily identify where to find these solutions.

> as suggested on purescript.org

That's a good point. Perhaps link to these same troubleshooting tips from there too.

hdgarrood:
I don’t think a separate section called “troubleshooting” or “FAQ” should be created in INSTALL.md, since the focus of the majority of that file is already to help people install the compiler or fix common installation issues. I think including the EACCES error in there too would be a good idea.

guides/Getting-Started.md
```diff
@@ -4,9 +4,9 @@ Let's walk through the basics of getting set up to use the PureScript compiler `
...
 
-#### Installing the Compiler
+### Installing the Compiler
 
-You'll need to install [Node.js and npm](https://docs.npmjs.com/getting-started/installing-node).  We recommend installing [Node.js and npm via a node version manager](https://docs.npmjs.com/getting-started/installing-node) to avoid issues with installing packages globally. If you choose to install it manually, you might experience the [`EACCES` error when installing packages globally](https://docs.npmjs.com/getting-started/fixing-npm-permissions#option-1-change-the-permission-to-npm-s-default-directory).
+You'll need to install [Node.js and npm](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#using-a-node-version-manager-to-install-nodejs-and-npm).  We recommend installing [Node.js and npm via a node version manager](https://docs.npmjs.com/getting-started/installing-node) to avoid issues with installing packages globally. If you choose to install it manually, you might experience the [`EACCES` error when installing packages globally](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
```
thomashoneyman, line 9: Perhaps this could say "We suggest installing the compiler with NPM, though other installation methods exist" and link over to the INSTALL.md file?

milesfrain: I originally went with the minimal-change strategy for this section by just adding the correct anchors, but will go ahead with my interpretation of your suggestion.

milesfrain:
Applied latest review feedback and also simplified the installation instructions by linking to the compiler's [INSTALL.md](https://github.com/purescript/purescript/blob/master/INSTALL.md).

The removed content is being reintroduced in https://github.com/purescript/purescript/pull/3920

hdgarrood (approved):

PR #324, opened 2020-06-11
Proposed corrections and changes to getting started guide

oldfartdeveloper:
@milesfrain in attempting to reconcile Chapter 8's procedures w/ your explanation of how to do it, I combed through the _Getting Started_ guide.  In the process, I found some mistakes and added what I thought are a couple of desired missing details.

One judgement call I made was to have GitHub _not_ track the user's `my-project` directory.

Hope this helps.  I'll drop duplicated information in the chapter 8 document.


guides/Getting-Started.md
```diff
@@ -258,15 +262,39 @@ Spago can be used to turn our PureScript code into JavaScript suitable for use i
...
+
+</html>
+```
+
+See this HTML file work as follows:
+
+1.  Run `npx parcel index.html --open` in a terminal opened to your `my-project`.
+1.  In the newly-opened browser, browse `localhost:1234`.  You will probably see a blank window.
+1.  Open a **debugger console**.
+1.  Observe the console contains the Euler exercise result string `The answer is 233168`.
```
oldfartdeveloper, line 289: @milesfrain At this point, the transition to the explanation of the `index.js` file is abrupt (my doing).  What occurs to me is I'd like to see adding instructions to write the Euler result to a text control that shows in the HTML page.  Then provide the smooth transition to the `index.js` file only explain its sending the result to the **HTML text control** rather than to the **console**.  Would this be worthwhile?

milesfrain: I'd support adding a "Rendering in Browser" section after this one, that shows how to render the answer as HTML text to the page and describing why `spago run` breaks.

.gitignore
```diff
@@ -0,0 +1 @@
+/my-project
```
milesfrain, line 1: I don't think we need this file here

guides/Getting-Started.md
```diff
@@ -77,6 +77,7 @@ Open PSCi by typing `spago repl` at the command line. Optionally, you can create
 
 As the introduction indicates, you can type `:?` to see a list of commands:
 
+    ```
```
milesfrain, line 80: The indentation lets this be a code block, so the backtics aren't necessary. The rest of the file only uses them for language blocks.

milesfrain:
Looks good in general. Just a few comments.

Regarding adding a section on rendering to browser, perhaps let that be a separate PR, since that may involve some additional discussion beyond these straightforward corrections.

.gitignore
oldfartdeveloper: True, though I found it convenient to make sure _I_ didn't inadvertently check in `my-project`.

oldfartdeveloper:
I'll make your corrections, thanks.

In addition, per our [discussion here](https://github.com/purescript/documentation/pull/324#issuecomment-642985881), I'll add a PR to the code itself to use the `diagonal` exercise instead of `Euler`.

oldfartdeveloper:
@milesfrain all of your suggestions have been handled; this should be ready to merge.

guides/Getting-Started.md
```diff
@@ -258,15 +260,41 @@ Spago can be used to turn our PureScript code into JavaScript suitable for use i
...
+See this HTML file work as follows:
+
+1.  Run `npx parcel index.html --open` in a terminal opened to your `my-project`.
+1.  In the newly-opened browser, browse `localhost:1234`.  You will probably see a blank window.
+1.  Open a **debugger console**.
+1.  Observe the console contains the Euler exercise result string `The answer is 233168`.
 
 If you open `index.js`, you should see a few compiled modules which look like this:
 
+#### How Purescript Compiler Generates Javascript
```
hdgarrood, line 291: Nit: It's "PureScript", not "Purescript". Could you move this header up above the "If you open index.js" line too please?

guides/Getting-Started.md
```diff
@@ -258,15 +260,41 @@ Spago can be used to turn our PureScript code into JavaScript suitable for use i
...
+  <div id="container"></div>
+  <script src="./index.js"></script>
+</body>
+
+</html>
+```
+
+See this HTML file work as follows:
+
+1.  Run `npx parcel index.html --open` in a terminal opened to your `my-project`.
```
hdgarrood, line 284: I'd rather not involve npx or parcel at this stage. Why not ask the reader to open index.html in a browser directly (i.e. with a file:// url)? Some projects need to use an HTTP server for local development because of things like cookies, but that shouldn't be necessary here.

guides/Getting-Started.md
hdgarrood: I'd rather not, actually; I think this should be beyond the scope of this guide, especially because there are quite a few options for this, and none of them seem ideal. I don't think the low-level DOM API bindings are a good fit because they're very verbose and people are unlikely to be using them in a real project either; alternatively you could use one of the web UI libraries like halogen or react-basic, but then you have to privilege one library over the others, and I think such documentation is better placed in those individual libraries anyway.

guides/Getting-Started.md
```diff
@@ -258,15 +260,41 @@ Spago can be used to turn our PureScript code into JavaScript suitable for use i
...
+1.  In the newly-opened browser, browse `localhost:1234`.  You will probably see a blank window.
+1.  Open a **debugger console**.
```
hdgarrood, line 286: I don't really like the bold here, could this just be normal text please?

guides/Getting-Started.md
```diff
@@ -258,15 +260,41 @@ Spago can be used to turn our PureScript code into JavaScript suitable for use i
...
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>Euler Exercise</title>
+</head>
+
+<body>
+  <div id="container"></div>
```
hdgarrood, line 275: I don't think this div is doing anything, so could we remove it?

guides/Getting-Started.md
hdgarrood: Ah, I'd forgot you had opened an issue for this already (#325), let's have that discussion there then.

guides/Getting-Started.md
oldfartdeveloper: @hdgarrood there were two instances of `Purescript` in this doc; updated both.  Good catch on moving `index.js`.

hdgarrood: Thanks! Oh sorry I also just spotted this, could you add a “the” before “PureScript” too please?

guides/Getting-Started.md
oldfartdeveloper: @hdgarrood Without a server, opening the debugger console doesn't show the "The answer is 233168" line (at least in my system).  But you have more important concerns below.

hdgarrood: That's weird, it does for me. Does your index.js end with `PS["Main"].main();`? What browser are you using?

guides/Getting-Started.md
oldfartdeveloper: Yep, done (and you'll see it when I upload it).

guides/Getting-Started.md
oldfartdeveloper: Yup, done.

guides/Getting-Started.md
oldfartdeveloper: Removed.

guides/Getting-Started.md
oldfartdeveloper: @hdgarrood my bad; it worked after I asked my FireFox browser to "show content messages."

thomashoneyman: I agree with @hdgarrood -- since we can just open this in the browser via the `index.html` file we may as well only require that vs. introducing a particular JavaScript bundler. I think we should only introduce the JS tools when they become necessary as they're outside our ecosystem.

guides/Getting-Started.md
thomashoneyman: I think it's fine to say that if you use the browser to run your PureScript code that output will be written to the browser console instead of being written to your terminal. I agree that this minimal getting started guide shouldn't introduce any UI libraries or use the DOM bindings.

JordanMartinez:
This PR hasn't been merged because it now seems a bit outdated

oldfartdeveloper:
@JordanMartinez I'm reviewing what has been accomplished thus far.  If my PR is still useful, I'll make the necessary changes; otherwise I'll just close this one.  Thanks for the heads-up!

oldfartdeveloper:
@JordanMartinez I've converted the `parcel` instructions to simply open the `Getting-Started` document with a browser and examine the Euler results in the debugger.

I believe this satisfies the objectives by @thomashoneyman, @hdgarrood, and @milesfrain.

oldfartdeveloper:
@JordanMartinez you are correct; the issues I have addressed here are already in `master` on Dec 29, 2022.

Hence, closing this issue.  I'll submit a new issue/PR for new minor issues that I found while checking the latest release soon.

PR #330, opened 2020-06-28
Add section on record puns and type synonyms for records

milesfrain:


language/Records.md
```diff
@@ -20,6 +20,15 @@ Fields of records can be accessed using a dot, followed by the label of the fiel
...
 
+Type synonyms for record types are created with the `type` keyword:
+
+```purs
+type Point =
+  { x :: Number
+  , y :: Number
+  }
+```
+
```
thomashoneyman, line 31: This is covered later in this documentation, but I suppose it's information worth stating right away. Why not stick with the existing example in this section, though?

> It is common to use type synonyms to for records via the `type` keyword:
>
> ```purs
> type Author = { name :: String, interests :: Array String }
>
> author :: Author
> author = { name: "Phil", interests: ["Functional Programming", "JavaScript"] }
> ```

language/Records.md
```diff
@@ -79,6 +88,65 @@ A record update function can also be defined by using an `_` inplace of the reco
...
+Here's an example of using the `disjointUnion` function to add a `z` field to a `Point` record. The `to3d` function merges the original record `p` with another record `{ z }` created from this new `Number`:
+
+```purescript
+import Record (disjointUnion)
+
+type Point3d
+  = { x :: Number
+    , y :: Number
+    , z :: Number
+    }
```
thomashoneyman, line 137: I think typical syntax in this repository places the `=` on the same line as the `type` keyword:

```purs
type Point3d =
  { x :: Number
  , y :: Number
  , z :: Number
  }
```

language/Records.md
```diff
@@ -79,6 +88,65 @@ A record update function can also be defined by using an `_` inplace of the reco
...
+
+to3d :: Point -> Number -> Point3d
+to3d p z = disjointUnion p { z }
+
+-- Equivalent to:
+to3d p z =
```
thomashoneyman, line 143: We could probably put this all on one line

language/Records.md
```diff
@@ -79,6 +88,65 @@ A record update function can also be defined by using an `_` inplace of the reco
 _ { fieldName = newValue }
 ```
 
+## Record Puns
```
thomashoneyman, line 91: I like this section, but there's some overlap with the language syntax documentation:
https://github.com/purescript/documentation/blob/837ec4ff3a956cdeda31997bf5f697ee738295b6/language/Syntax.md#additional-forms-with-records

I think we should merge that content into this file and then update the language syntax file to link here. It makes more sense for that content to be in this section and just a link be in that file.

thomashoneyman: I think that file has good content and it could be used to inform what you've written here.

language/Records.md
```diff
@@ -79,6 +88,65 @@ A record update function can also be defined by using an `_` inplace of the reco
...
+
+Note that puns may not be used for record updates.
+
+```purescript
+setX :: Number -> Point -> Point
+setX x point = point { x = x }
+-- setX x point = point { x } -- Not allowed
+```
+
+## Merging Records
```
thomashoneyman, line 124: I'm glad to see the `record` library mentioned here. Perhaps this section could be named "Further record operations" and it could focus on merging but note that the `record` library provides other things too (for example, an efficient record builder via `Builder`).

thomashoneyman:
As a side note, I think the `Kinds` section in here could use a little extra clarity. Specifically, I'm thinking of something like:

---

`{ ... }` is syntactic sugar for the `Record` type constructor. This type constructor is parameterized by a row of types:

```purs
-- these types are equivalent
type Language = { language :: String }
type Language' = Record ( language :: String )
```

A `Record` is constructed from a row type and represents a product type in which all fields in the row type are present. Using kind notation, `Record` has the kind `# Type -> Type` -- that is, it takes a row of types and produces a type.

Because `( language :: String )` denotes a row of types (and therefore has the kind `# Type`), it can be passed to the `Record` constructor or to the `{ ... }` syntax for `Record` to construct a type:

```purs
type LanguageRow = ( language :: String ) -- has kind # Type

-- these are equivalent
type Language = Record LanguageRow
type Language' = { | LanguageRow }
```

language/Records.md
milesfrain: Applied suggestion

language/Records.md
milesfrain: Applied suggestions of changing title and more explicitly linking to the `Record.Builder` module.

milesfrain:
Applied all suggestions except for moving content from `Syntax.md` to `Records.md`. I think that would still be good to do as a separate PR. Opened https://github.com/purescript/documentation/issues/368 to follow-up.

JordanMartinez:
This PR hasn't been merged because this is likely useful, but has a few additions that are undesirable / repetitive

PR #333, opened 2020-07-16
Add more CycleInDeclaration examples

milesfrain:
I found the `x = x` example a bit too minimal to figure out what's required to fix these errors.

### DIFF
--- a/errors/CycleInDeclaration.md
+++ b/errors/CycleInDeclaration.md
@@ -39,3 +39,45 @@ Note that cycles can also spring up in much less obvious ways, e.g. if you defin
 ## Notes
 
 - If you are trying to implement a recursive parser in PureScript, then you might find this post on [recursive parsing in PureScript](https://github.com/Thimoteus/SandScript/wiki/2.-Parsing-recursively) to be helpful.
+
+### Additional Examples
+
+Mutually recursive functions:
+```purs
+{-
+A very inneficient way to write:
+zeroOrNine n = if n >= 9 then 9 else 0
+-}
+zeroOrNine :: Int -> Int
+zeroOrNine 9 = 9
+zeroOrNine n | n < 0 = 0
+zeroOrNine n = again (n - 1)
+
+again :: Int -> Int
+again = zeroOrNine
+{-
+Fix compiler error by replacing the above line with either:
+again n = zeroOrNine n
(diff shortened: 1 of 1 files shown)


errors/CycleInDeclaration.md
```diff
@@ -39,3 +39,45 @@ Note that cycles can also spring up in much less obvious ways, e.g. if you defin
...
+-- This builds, but blows-up call stack and triggers a
+-- "too much recursion error".
+instance showChain :: Show a => Show (Chain a) where
+  show = genericShow
+{-
+Fix the issue by replacing the above line with:
+  show c = genericShow c
+-}
+```
+https://try.purescript.org/?gist=32b78ce065d60427620cdd8d40777a1f
```
thomashoneyman, line 83: Perhaps this could just be a link to the contents being added in #338?

errors/CycleInDeclaration.md
```diff
@@ -39,3 +39,45 @@ Note that cycles can also spring up in much less obvious ways, e.g. if you defin
 ## Notes
 
 - If you are trying to implement a recursive parser in PureScript, then you might find this post on [recursive parsing in PureScript](https://github.com/Thimoteus/SandScript/wiki/2.-Parsing-recursively) to be helpful.
+
+### Additional Examples
+
+Mutually recursive functions:
```
thomashoneyman, line 45: If we move the generic deriving example to be a link then we can change this header to be `### Mutual recursion` and just focus on this mutual recursion example. I think that fits better with how [other errors use the Notes section](https://github.com/purescript/documentation/blob/2059f1d272aa863c465bcd4b0ade815a0bdfccc4/errors/KindsDoNotUnify.md).

Also, what do you think of trimming the example a little bit to something like this?

```purs
isEven :: Int -> Boolean
isEven n 
  | n == 0 = true
  | otherwise = again (n - 1)

again :: Int -> Boolean
again = isEven
```

which also fails, producing this error:

```
  The value of isEven is undefined here, so this reference is not allowed.

while checking that expression isEven
  has type Int -> Boolean
in binding group isEven, again
```

JordanMartinez:
This PR hasn't been merged because I'm not sure that should be in the Getting Started section if it's in reference to the Book. Shouldn't the Book clarify that instead?

PR #331, opened 2020-07-10
Add derive via Generic example, and point to latest docs

milesfrain:
I often visit this page looking for a `genericShow` example.

It was also frustrating to click on the `Data.Generic.Rep` and be taken to an older package version that's missing lots of nice documentation. So ensuring links now go to the latest version. (related to https://github.com/purescript/pursuit/issues/414)

PR #338, opened 2020-07-21
Add more Type Class deriving details

### LINKED ISSUE #155: Document `derive newtype` syntax and capability
I can't find documentation for `derive newtype` syntax in this repo, and it should be added.

milesfrain:
From today's conversation on Slack.

Fixes #155

I think there are better ways to organize this content, but wanted to track this first version.

### DIFF
--- a/language/Type-Classes.md
+++ b/language/Type-Classes.md
@@ -199,6 +199,10 @@ derive instance eqScore :: Eq Score
 derive newtype instance eqScore :: Eq Score
 ```
 
+### Deriving from `Generic`
+
+The compiler's built-in support for `Generic` unlocks convenient deriving for many other classes not listed above. See the [deriving guide](../guides/Type-Class-Deriving.md) for more information.
+
 ## Compiler-Solvable Type Classes
 
 Some type classes can be automatically solved by the PureScript Compiler without requiring you place a PureScript statement, like `derive instance`, in your source code.
--- a/language/Type-Classes.md
+++ b/language/Type-Classes.md
@@ -137,11 +137,38 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
 
 ## Type Class Deriving
 
-The compiler can derive type class instances to spare you the tedium of writing boilerplate. There are a few ways to do this depending on the specific Type and Class being derived.
+The compiler can derive type class instances to spare you the tedium of writing boilerplate. There are a few ways to do this depending on the specific type and class being derived.
+
+### Classes with built-in compiler support
+
+Some classes have special built-in compiler support, and their instances can be derived from all types.
+
+For example, if you you'd like to be able to remove duplicates from an array of an ADT using `nub`, you need an `Eq` and `Ord` instance. Rather than writing these manually, let the compiler do the work.
+
+```purs
+import Data.Array (nub)
+
+data MyADT
+  = Some
+  | Arbitrary Int
+  | Contents Number String
+
+derive instance eqMyADT :: Eq MyADT
+derive instance ordMyADT :: Ord MyADT
+
+nub [Some, Arbitrary 1, Some, Some] == [Some, Arbitrary 1]
+```
+
+Currently, instances for the following classes can be derived by the compiler:
+- [Data.Generic.Rep (class Generic)](https://pursuit.purescript.org/packages/purescript-generics-rep/docs/Data.Generic.Rep#t:Generic)
+- [Data.Eq (class Eq)](https://pursuit.purescript.org/packages/purescript-prelude/docs/Data.Eq#t:Eq)
+- [Data.Ord (class Ord)](https://pursuit.purescript.org/packages/purescript-prelude/docs/Data.Ord#t:Ord)
+- [Data.Functor (class Functor)](https://pursuit.purescript.org/packages/purescript-prelude/docs/Data.Functor#t:Functor)
+- [Data.Newtype (class Newtype)](https://pursuit.purescript.org/packages/purescript-newtype/docs/Data.Newtype#t:Newtype)
 
 ### Derive from `newtype`
 
-If you'd like to use a wrapped type's instance for your newtype, use the `derive newtype` keywords.
+If you would like your newtype to defer to the instance that the underlying type uses for a given class, then you can use newtype deriving via the `derive newtype` keywords.
 
 For example, let's say you want to add two `Points` values using the `Semiring` instance of the wrapped `Int`.
 
@@ -165,34 +192,7 @@ instance semiringPoints :: Semiring Points where
   one = Points 1
 ```
 
-### Classes with built-in compiler support
-
-Some classes have special built-in compiler support, and instances can be derived from all types, not just `newtype`s.
-
-For example, if you you'd like to be able to remove duplicates from an array of an ADT using `nub`, you need an `Eq` and `Ord` instance. Rather than writing these manually, let the compiler do the work.
-
-```purs
-import Data.Array (nub)
-
-data MyADT
-  = Some
-  | Arbitrary Int
-  | Contents Number String
-
-derive instance eqMyADT :: Eq MyADT
-derive instance ordMyADT :: Ord MyADT
-
-nub [Some, Arbitrary 1, Some, Some] == [Some, Arbitrary 1]
-```
-
-Currently, the following type classes can be derived by the compiler:
-- [Data.Generic.Rep (class Generic)](https://pursuit.purescript.org/packages/purescript-generics-rep/docs/Data.Generic.Rep#t:Generic)
-- [Data.Eq (class Eq)](https://pursuit.purescript.org/packages/purescript-prelude/docs/Data.Eq#t:Eq)
-- [Data.Ord (class Ord)](https://pursuit.purescript.org/packages/purescript-prelude/docs/Data.Ord#t:Ord)
-- [Data.Functor (class Functor)](https://pursuit.purescript.org/packages/purescript-prelude/docs/Data.Functor#t:Functor)
-- [Data.Newtype (class Newtype)](https://pursuit.purescript.org/packages/purescript-newtype/docs/Data.Newtype#t:Newtype)
-
-Note that with our earlier `Points` `newtype`, we can use either of these options to derive an `Eq` instance. They are equivalent in this case.
+Note that we can use either of these options to derive an `Eq` instance for a `newtype`, since `Eq` has built-in compiler support. They are equivalent in this case.
 
 ```purs
 derive instance eqPoints :: Eq Points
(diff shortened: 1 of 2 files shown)


Thimoteus:
Would this be an appropriate place to put info about eta abstraction to help fix some stack overflow issues?

milesfrain:
Yeah. That seems like a good addition.
Could include something similar to: https://github.com/purescript-contrib/purescript-argonaut-codecs/blob/master/README.md#deriving-instances
Might also be good to link to https://github.com/purescript-contrib/purescript-argonaut-generic for an example of writing generic support for your own classes.

milesfrain:
This should be ready to merge now (pending any additional feedback).

language/Type-Classes.md
```diff
@@ -137,36 +137,128 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
 
 ## Type Class Deriving
 
-Some type class instances can be derived automatically by the PureScript compiler. To derive a type class instance, use the `derive instance` keywords:
+The compiler can derive type class instances to spare you the tedium of writing boilerplate. There are a few ways to do this depending on the specific Type and Class being derived.
```
thomashoneyman, line 140: I think we typically leave `type` and `class` lower-cased.

language/Type-Classes.md
```diff
@@ -137,36 +137,128 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
 
-```purescript
-newtype Person = Person { name :: String, age :: Int }
+### Derive from `newtype`
+
+If you'd like to use a wrapped type's instance for your newtype, use the `derive newtype` keywords.
```
thomashoneyman, line 144: I'm not sure this fully captures the idea that you're deferring to the wrapped type's instance -- that it's almost like your newtype is ignored and the instance goes directly to whatever the wrapped type does for that class.

What about this tweak?

> If you would like your newtype to defer to the instance that the underlying type uses for a given class, then you can use newtype deriving via the `derive newtype` keywords.

language/Type-Classes.md
```diff
@@ -137,36 +137,128 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
 
 ## Type Class Deriving
 
-Some type class instances can be derived automatically by the PureScript compiler. To derive a type class instance, use the `derive instance` keywords:
+The compiler can derive type class instances to spare you the tedium of writing boilerplate. There are a few ways to do this depending on the specific Type and Class being derived.
 
-```purescript
-newtype Person = Person { name :: String, age :: Int }
+### Derive from `newtype`
```
thomashoneyman, line 142: I think newtype deriving should come after regular deriving. It's really common to throw together a sum type and then derive `Eq` and `Ord` instances, for example, and I think that's a nice motivating example for deriving in general.

Newtype deriving is a bit special because a) you can derive more things and b) it has some subtle behavior around instances using the same implementation as the underlying type that deserves an explanation.

What do you think of shifting this to after `derive` has been introduced?

language/Type-Classes.md
```diff
@@ -137,36 +137,128 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
+
+If you'd like to use a wrapped type's instance for your newtype, use the `derive newtype` keywords.
+
+For example, let's say you want to add two `Points` values using the `Semiring` instance of the wrapped `Int`.
```
thomashoneyman, line 146: I think the `semiring` example is a great example of something useful that (iirc) can't be derived via the usual `derive` keyword and has to use newtype deriving. However, the example that I think best illustrates what exactly newtype deriving is doing is `Show`.

Specifically:

```purs
newtype Point = Point Int

derive newtype instance showPoint :: Show Point
```

and then in the repl:

```
> show (Point 1)
"1"
```

This illustrates that you're really using the show instance for `Int` -- you don't see the "Point" constructor at all. Same idea applies to things like deriving a `DecodeJson` instance (which I don't think we should cover here).

I don't think that we need to delete the semiring example, but maybe we could also touch on the show example to demonstrate this further.

language/Type-Classes.md
```diff
@@ -137,36 +137,128 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
+derive newtype instance eqPoints :: Eq Points
+```
+
+### Deriving from `Generic`
+
+The compiler's built-in support for `Generic` unlocks convenient deriving for many other classes not listed above.
+
+For example, if we wanted a `Show` instance for `MyADT`, it might seem like we're out of luck because neither `MyADT` is a `newtype` nor is `Show` listed as a class with built-in compiler support.
+
+But we _can_ use `genricShow`, which works with _any_ type that has a `Generic` instance. And recall that the compiler has built-in support for deriving a `Generic` instance for any type (including the `MyADT` type). We put all those pieces together like so:
```
thomashoneyman, line 208:
```suggestion
But we _can_ use `genericShow`, which works with _any_ type that has a `Generic` instance. And recall that the compiler has built-in support for deriving a `Generic` instance for any type (including the `MyADT` type). We put all those pieces together like so:
```

This might be the better place to demonstrate the `Show` differences between newtype deriving and generic deriving, actually, rather than earlier. I believe that the generic version of `Points` will include the constructor and the newtype derived version will not.

language/Type-Classes.md
```diff
@@ -137,36 +137,128 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
+main = logShow [Some, Arbitrary 1, Contents 2.0 "Three"]
+-- Prints:
+-- [Some,(Arbitrary 1),(Contents 2.0 "Three")]
 ```
 
 More information on Generic deriving is available [in the generics-rep library documentation](https://pursuit.purescript.org/packages/purescript-generics-rep).
 
+#### Avoiding stack overflow errors with recursive types
+
+Be careful when using generic functions with recursive data types. The instances _cannot_ be written in point free style:
```
thomashoneyman, line 229:
```suggestion
Be careful when using generic functions with recursive data types. Due to strictness, these instances _cannot_ be written in point free style:
```

language/Type-Classes.md
milesfrain: > you don't see the "Point" constructor at all.

This is clearer to contrast if the reader is already familiar with `generic`, which I believe is the only `derive` option that prints the constructor. I can still include a lookahead note. Wondering if there's another example that doesn't depend on `generic`.

language/Type-Classes.md
milesfrain: Geeze. I should read all the comments before commenting myself.

milesfrain:
Applied all of @thomashoneyman's suggestions

andys8 (approved):

language/Type-Classes.md
```diff
@@ -137,35 +137,159 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
+```purs
+derive instance eqScore :: Eq Score
+derive newtype instance eqScore :: Eq Score
+```
+
+### Deriving from `Generic`
+
+The compiler's built-in support for `Generic` unlocks convenient deriving for many other classes not listed above.
+
+For example, if we wanted a `Show` instance for `MyADT`, it might seem like we're out of luck because neither `MyADT` is a `newtype` nor is `Show` listed as a class with built-in compiler support.
```
thomashoneyman, line 206: Minor grammar typo with `neither` here, but GitHub is not letting me add a suggestion so I'm copy/pasting a suggestion:

> For example, if we wanted a `Show` instance for `MyADT` it might seem like we're out of luck:  `Show` is not a class with built-in compiler support for deriving and `MyADT` is not a `newtype` (so we can't use newtype deriving).

language/Type-Classes.md
```diff
@@ -137,35 +137,159 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
-instance showPerson :: Show Person where
+instance showMyADT :: Show MyADT where
   show = genericShow
+  
+main = logShow [Some, Arbitrary 1, Contents 2.0 "Three"]
+-- Prints:
+-- [Some,(Arbitrary 1),(Contents 2.0 "Three")]
+```
+
+Note that the `show` output string can be copy-pasted to reconstruct the original data. This is a good oportunity to emphasize how `newtype` deriving simply reuses the instance of the underlying type. In the case of `show`, this means omitting printing the wrapper.
```
thomashoneyman, line 225:
```suggestion
The `Show` type class is most often used for debugging data, so the output of most `Show` instances can be copy-pasted back into a PureScript source file to reconstruct the original data. The `Show` instance we created by deriving `Generic` and then using `genericShow` follows this convention.

This is a good opportunity to emphasize how newtype deriving is different from instances derived by the compiler or through the `Generic` type class. In the examples below, notice how the instance derived through `Generic` includes the newtype constructor `Score`, but the newtype-derived instance simply reuses the underlying `Show` instance for `Int` and therefore does not include the constructor:
```

language/Type-Classes.md
```diff
@@ -137,35 +137,159 @@ See also the section in [PureScript by Example](https://book.purescript.org/chap
...
+```purs
+newtype Score = Score Int
+
+derive newtype instance semiringScore :: Semiring Score
+
+tenPoints :: Score
+tenPoints = (Score 4) + (Score 6)
+```
+
+That `derive` line replaced all this code!:
```
thomashoneyman, line 184:
```suggestion
That `derive` line replaced all this code:
```

milesfrain: We'll just have to let readers form their own conclusions about how exciting this feature is.

hdgarrood:
This is good, but the stuff beyond Generic is not suitable for the language reference, as it's more of a how-to/explanation and covers things which aren't part of the language.

milesfrain:
Section removed. Hoping to include that in some sort of language "guide" in the future.

hdgarrood:
Thanks. I think it would be good to put those other paragraphs into the “guides” directory instead. Even if there’s a bit of overlap between the reference and the guide, I think that’s fine.

hdgarrood:
Oh sorry; to clarify, I think the stuff under the “Deriving with Generic” heading is also not really suitable for the language reference.

thomashoneyman:
I agree the content deserves a place in the documentation (rather than being removed altogether) and perhaps we could link to it here as further reading in the guides section.

thomashoneyman (approved):
I'm happy with this content ending up in the guides section and linked from the language reference.

milesfrain:
The first two sections (built-in and newtype) are duplicated across both the guide and the language reference. I feel like this content is equally relevant for the guide, since the generic section builds on these. There are certainly other ways to organize this material, such as:
1. Relocating just the `Generic` instructions to `generics-rep` as proposed in https://github.com/purescript/purescript-generics-rep/issues/41
1. Linking back to the prerequisite sections of the language reference
1. Deleting the deriving guide in this docs repo

But I'm thinking it's best to save this reorganization work for later.

My understanding is that the `language` folder is supposed to be as concise as possible. **Is this content considered the language "specification", "reference", or both?** Some languages use those terms interchangeably https://golang.org/ref/spec

I'm still unsure what's considered within scope of these sections. In hindsight, it seems like the `Generic` example merged in #331 was scope creep.

hdgarrood:
The duplication is not a problem in my mind. I’d much rather have a little bit of duplication here and there than have documentation resources which try to cater to everyone at once.

The `language/` directory is more of a reference than a specification. In my mind a specification ought to tell you how everything behaves in a much more precise manner, and is aimed more at implementers (whereas a reference is aimed at users). It would be good to have a specification one day (which would be another separate resource), but realistically that’s unlikely to happen any time soon.

I agree that the example merged in #331 probably didn’t belong in the language reference either, in retrospect. The main criterion is whether something is part of the language or whether it is part of a library; if it’s the latter, it probably shouldn’t go in here. Of course the boundaries aren’t always clear: it’s difficult to discuss do notation without mentioning monads, for example.

hdgarrood (approved):

PR #339, opened 2020-07-21
Add example of using guards with case expression

milesfrain:
This might not be the ideal example, but there should be some note on this combined usage.

### DIFF
--- a/language/Pattern-Matching.md
+++ b/language/Pattern-Matching.md
@@ -176,6 +176,30 @@ compare _ _ = EQ
 
 (The name `otherwise` is a synonym for `true` commonly used in guards.)
 
+Guards may also be used within `case` expressions, which allow for inline expressions. For example, these are equivalent: 
+
+```purs
+fb :: Int -> Effect Unit
+fb n =
+  log case n of
+    _
+      | 0 == mod n 15 -> "FizzBuzz"
+      | 0 == mod n 3 -> "Fizz"
+      | 0 == mod n 5 -> "Buzz"
+      | otherwise -> show n
+```
+
+```purs
+fb :: Int -> Effect Unit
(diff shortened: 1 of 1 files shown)


language/Pattern-Matching.md
```diff
@@ -176,6 +176,30 @@ compare _ _ = EQ
...
+```purs
+fb :: Int -> Effect Unit
+fb n =
+  log case n of
+    _
+      | 0 == mod n 15 -> "FizzBuzz"
+      | 0 == mod n 3 -> "Fizz"
+      | 0 == mod n 5 -> "Buzz"
+      | otherwise -> show n
+```
```
thomashoneyman, line 190: I'm happy with the example. I think I most commonly see this done with a `case _ of` in the wild, as the variable name is only used once then and closer to the guards. What about this simplified version?

```purs
fizzBuzz :: Int -> String
fizzBuzz = case _ of
  n
    | 0 == mod n 15 -> "FizzBuzz"
    | 0 == mod n 3 -> "Fizz"
    | 0 == mod n 5 -> "Buzz
    | otherwise -> show n
```

or if you want to stay in `Effect` we can do `fizzBuzz = log <<< case _ of`

hdgarrood:
Thanks! I think this example could be improved by making use of the ability to pattern match as well as using a guard in `case`, with something like this, as these are the cases when it comes in useful:

```purescript
case x of
  Just y | someCondition y -> ...
  Just y -> ...
  Nothing -> ...
```

I think it's a little bit odd to write a `case` expression like this one where you only have one case alternative so you could just use guards instead (as you have in the version that you've pointed out that it's equivalent to). I agree though that this is fine to merge as-is, of course, since I've merged it.

Issue #274, opened 2019-09-26
Add recommended tooling for PureScript applications


thomashoneyman:
I recently posted a short Discourse thread, [Recommended Tooling for PureScript Applications in 2019](https://discourse.purescript.org/t/recommended-tooling-for-purescript-applications-in-2019). It clears up the tools that I recommend when new folks ask what they're supposed to use to build their PureScript web applications. It's received a warm response in Slack and on Discourse.

It's hard enough getting started with a new language and it helps beginners to know what they can at least install to get a new single-page application going. At minimum, I'm glad to see things moving towards recommending Spago, but it would also be useful to recommend Yarn for npm packages / script running and Parcel or Webpack for bundling. I think these are the three categories of tool that are necessary to build a typical PureScript app which involves any JavaScript libraries.

Note that this would only be about single-page applications; PureScript is used in other contexts, too, and in those contexts Yarn and Parcel don't necessarily make sense to recommend.

To reiterate: I suggest updating the documentation repository and/or purescript.org to recommend Yarn, Spago, and Parcel (or Webpack) for building web applications with PureScript, while also noting that other tools exist and are perfectly usable.

thomashoneyman:
Alternately we could just crib parts of the post itself into the documentation, but the post is probably too opinionated and touches on too many tools for that.

hdgarrood:
Why Yarn over npm? npm does more or less everything I want it to, and it comes with node.js, which is a big plus. I’d prefer to let people discover and use Yarn themselves if they aren’t happy with npm, personally.

thomashoneyman:
Now that I'm looking around it looks like npm has improved dramatically since the old days of non-deterministic node_modules directories and slow installation times. With that in mind I don't have strong arguments for one or the other, and I understand that `npm` already coming with the node installation is a plus.

I'm not an npm user myself these days so I'm not up-to-date on arguments for or against and I'll have to defer to others more knowledgeable than I on the topic.

There was also [Connor Prussin’s comment on npx](https://discourse.purescript.org/t/recommended-tooling-for-purescript-applications-in-2019/948/3?u=thomashoneyman) in the tooling thread, though it was directed at yarn + npx.

With all this in mind, I'd also be happy with npm + spago + parcel/webpack.

hdgarrood:
[...]

PR #343, opened 2020-08-16
Suggest resources

toraritte:
Replaced the "PureScript and NixOS" post with `easy-purescript-nix` as the former post is outdated (2015) and the latter is a widely accepted tool that can be used out of the box.

Thank you for considering!


README.md
```diff
@@ -38,14 +39,15 @@ This repository is a collaborative effort, so please feel free to make a pull re
...
-- [PureScript and NixOS](https://pr06lefs.wordpress.com/2015/01/11/get-started-with-purescript-on-nixos/): How to use PureScript with NixOS
+- [`easy-purescript-nix`](https://github.com/justinwoo/easy-purescript-nix): Set up PureScript and development environment with [Nix](https://nixos.org/)
 - [PSCi](guides/PSCi.md): An interactive development tool for PureScript
 
 ### Ecosystem
 
 - [Maintained Packages](ecosystem/Maintained-Packages.md)
 - [Style Guide](guides/Style-Guide.md)
 - [Alternate Backends](https://github.com/purescript/documentation/blob/master/ecosystem/Alternate-backends.md): PureScript can compile to other languages as well!
+- [Recommended tooling for PureScript in 2020](https://discourse.purescript.org/t/recommended-tooling-for-purescript-in-2020/1615)
```
milesfrain, line 50: There's an attempt to do this for the 2019 thread in https://github.com/purescript/documentation/issues/274. Some of the considerations discussed there also apply to this PR. I'm personally in favor of linking to that thread, or even tracking its contents in this repo so it can more easily be continuously updated.

thomashoneyman: I think this and the above reference to `easy-purescript-nix` can be removed once #343 is merged, as they will both be included in the "Editor and tool support" documentation.

PR #347, opened 2020-08-21
Update editor and tooling support

### LINKED ISSUE #274: Add recommended tooling for PureScript applications

thomashoneyman:
Relates to #343 and closes #274. This merges the contents from the [Recommended Tooling for PureScript Applications in 2020](https://discourse.purescript.org/t/recommended-tooling-for-purescript-in-2020) post on Discourse into the editor and tooling support post.

It's somewhat of an overhaul of the existing documentation page, but I believe the new version accomplishes a few goals:

1. It makes it clear right away what the minimum recommended tools are for a typical PureScript project
2. It emphasizes Spago as the default choice for package management and build tooling
3. It includes tools from the PureScript ecosystem which are useful for developing libraries and applications that are not only for use with editors.
4. It mentions necessary tools when integrating with another ecosystem for a particular backend (namely, JavaScript tools for projects which integrate with JS).

This documentation could be narrowed to _only_ include build tooling and editor tooling, but I think it's sensible to include things like setup-purescript, zephyr, and purty here -- they're useful for development generally and there's not really a place for them elsewhere in the documentation at the moment. We could move them elsewhere later on.

I moved the `gulp-purescript` task to the deprecated section because that tool has been deprecated.

### DIFF
--- a/ecosystem/Editor-and-tool-support.md
+++ b/ecosystem/Editor-and-tool-support.md
@@ -11,7 +11,9 @@ These tools can be installed via [NPM](https://www.npmjs.com), among other insta
 
 ## Editor support
 
-The PureScript compiler includes an IDE server, [`psc-ide`](https://github.com/purescript/purescript/tree/master/psc-ide), to supply tooling for editors. Most editor plugins which rely on PureScript's IDE tooling have at least these features:
+The PureScript compiler includes an IDE server, [`psc-ide`](https://github.com/purescript/purescript/tree/master/psc-ide), to supply tooling for editors. This server has been used to implement a [PureScript language server](https://github.com/nwolverson/purescript-language-server), which implements the [Language Server Protocol](https://langserver.org) for PureScript. Major editor tooling either uses `psc-ide` or the language server.
+
+Most editor plugins which rely on PureScript's IDE tooling have at least these features:
 
 - Autocomplete (including auto-imports)
 - Definitions and error reporting on hover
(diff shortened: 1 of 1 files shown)


hdgarrood:
Looks great! I have a very minor nit that we should be avoiding "psc" now since the compiler executable no longer uses that name; I think the IDE server built into the compiler should ideally be referred to as `purs ide server` to minimise the risk of confusion. The editor plugins which include "psc-ide" in their names probably should be referred to by their actual names still, of course.

hdgarrood (approved):

ecosystem/Editor-and-tool-support.md
```diff
@@ -1,51 +1,111 @@
...
-Install these two packages and follow the instructions in the `psc-ide-emacs` README.
-- [purescript-mode](https://github.com/purescript-emacs/purescript-mode) Syntax Highlighting and indentation (adapted from haskell-mode)
-- [psc-ide-emacs](https://github.com/purescript-emacs/psc-ide-emacs) Emacs plugin exposing the compilers IDE support
+These tools can be installed via [NPM](https://www.npmjs.com), among other installation methods.
 
-PSCI support via comint:
+## Editor support
 
-- [psci-mode](https://github.com/purescript-emacs/emacs-psci) is a REPL minor mode
+The PureScript compiler includes an IDE server, [`psc-ide`](https://github.com/purescript/purescript/tree/master/psc-ide), to supply tooling for editors. This server has been used to implement a [PureScript language server](https://github.com/nwolverson/purescript-language-server), which implements the [Language Server Protocol](https://langserver.org) for PureScript. Major editor tooling either uses `psc-ide` or the language server.
```
hdgarrood, line 14: How about "Major editor tooling either uses `purs ide server` directly, or via the language server."?

Issue #345, opened 2020-08-19
Port Jordan's "Syntax" folder into this repo


JordanMartinez:
Here's what I propose: port [my entire Syntax Folder](https://github.com/JordanMartinez/purescript-jordans-reference/tree/latestRelease/11-Syntax) into this repo. (Note: the repo already has up-to-date syntax guide on the upcoming PS v0.14.0 release via the [v0.14.0-syntax](https://github.com/JordanMartinez/purescript-jordans-reference/tree/v0.14.0-syntax/11-Syntax) branch).

Here are the benefits:
- the entire language's syntax (at least as far as I know) is already documented in a very clear and readable way.
- the compiler is already being used to verify that the syntax is valid. As a result, it's easy to identify syntax that needs to be updated when breaking changes are released
- new learners will discover this folder's contents sooner and have a better learning process / new user experience

The costs:
- this repo seems intended primarily for a brief quick-start guide. Porting something this large into this repo could very well go "outside the scope of this project." (I can't recall if that's the reason why this hasn't been done before)
- my explanations would need to be reviewed. While most are true, I imagine people would bikeshed over the exact wording that _should_ be used to explain something as there very well might be nuances that I have failed to capture or incorrectly communicated.
- increased maintenance burden. The compiler would reduce this maintenance, but explanations would still take time to review. Moreover, each PS release that includes breaking changes would add this additional 'to-do' onto an already large list.

milesfrain:
I find the language folder in this docs repo (with a few .md files) much easier to consume than clicking though a tree of interleaved .purs and .md files.

I'd prioritize ease-of-reading over ease-of-maintaining, but getting both is ideal. Can we find a way to keep compiler verification, while still letting the docs be easy to read and navigate? Maybe something like [mdBook anchors](https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file)? Is this compatible with any of the ideas proposed in #344?

JordanMartinez:
> Can we find a way to keep compiler verification, while still letting the docs be easy to read and navigate? Maybe something like mdBook anchors?

Yeah, something like that is what I had in mind.

hdgarrood:
> It sounds like the language reference exists primarily for core contributors to the PureScript language and people already familiar with the language. It was my understanding that this repo was more for new beginners who want a quick overview of the syntax and other PureScript-related things. Is that clarified anywhere in this repo?

[...]

PR #348, opened 2020-08-24
Clarify intended audience of the language reference folder

JordanMartinez:
Per [Harry's comment about this](https://github.com/purescript/documentation/issues/345#issuecomment-678812487), this PR adds a note that the `language` folder's content is not intended for those learning PureScript for the first time.

I could also update the project's readme to clarify that in the [Learning section](https://github.com/purescript/documentation/#learning), but I'll wait for further guidance as to how to do that before doing the work. For example, should I include a note near that link that clarifies that? Should the link be relocated to somewhere else on this repo?

### DIFF
--- a/language/README.md
+++ b/language/README.md
@@ -1,5 +1,7 @@
 # PureScript Language Reference
 
+**Note:** the `language` folder's content is primarily for intermediate users already familiar with the language. It serves to help such users determine whether an issue they are experiencing is intended behavior or caused by a bug. It is not meant for people who are learning PureScript for the first time.
+
 As an introductory example, here is the usual "Hello World" written in PureScript:
 
 ```purescript


hdgarrood:
I'd like to think a bit more about how I'd like to present this: determining whether behaviour you're seeing is intended or not is just one reason that you might want to use this reference. You might also want to use this reference to double-check the specifics of a language feature if you don't completely remember how it works, or if you've seen some unusual syntax somewhere and you're not sure what it means. It should also be useful if you're trying to write tooling for the language: ideally I think it should be possible to write a parser which agrees with the compiler's parser using just the language reference, for example.

JordanMartinez:
Ok. Did you want to me to add what you've written above verbatim? Or should I close this PR and let you write something instead?

hdgarrood:
I don't really like the idea of adding what I've written above verbatim, because the intended audience of that comment was a fellow maintainer (you) whereas the intended audience of anything in this repo should be much wider. After having thought a bit, I also don't think we should try to direct beginners away from this resource quite as strongly as this PR does; it's true that a beginner is very unlikely to get very far with this resource _alone_, but if it's complementing something that's more of a guide or tutorial (e.g. the book) I think that's useful. For example, I was learning Rust recently and when I decided I wanted to learn things in a different order to how the book presents them, or just if I wanted to read up on a feature I'd just learned about in more detail, I found it very useful to jump between the Rust book and language reference.

Realistically though I'm unlikely to find the time to give this PR/issue the treatment it deserves, so don't wait for me to write something else.

JordanMartinez:
This never went anywhere, so I'm closing.

PR #349, opened 2020-08-27
Add note about IDE setup

milesfrain:
We should point this out to users. Not sure if this the best place for it.
Could alternatively do:
- Earlier in "Setting up the Development Environment"
- Later in "What Next?"


hdgarrood:
Putting it in "What next" sounds preferable to me, since we aren't doing anything that requires it in the guide, and people will be able to see something happen faster if they don't try to install editor tooling first. I think there's also a risk that having editor tooling running alongside the commands the guide asks you to perform might confuse things.

JordanMartinez:
This PR hasn't been merged because I'm not sure that should be in the Getting Started section if it's in reference to the Book. Shouldn't the Book clarify that instead?

PR #350, opened 2020-08-27
Note annotation requirements in let and where

milesfrain:
Inspired by https://github.com/purescript/purescript/issues/3922
This might be too nitty-gritty for initial reference, but I think this should be documented _somewhere_. If these details exist in another location, it should at least be discoverable from the `let`/`where` reference section.

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -527,6 +527,30 @@ factorial = go 1
   go acc n = go (acc * n) (n - 1)
 ```
 
+### Caveat for generalized functions
+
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think the language reference is the correct place to note this, but this isn’t to do with syntax, so I don’t think it should go in the Syntax.md file.

JordanMartinez:
This PR hasn't been merged because this PR is somewhat incorrect. The annotation is only needed when the same function is used at least twice and each usage takes an argument of a different type, such that the function must be polymorphic. Those rules aren't actually clarified in this PR.

PR #369, opened 2021-01-02
Add more examples for anonymous argument error

milesfrain:
Also modified the original examples to make them closer to what you'd find in real-world code.

I think this would be clearer if we split this up into multiple Example/Cause/Fix sections. Don't know if that deviation from the common format is allowed.

### DIFF
--- a/errors/IncorrectAnonymousArgument.md
+++ b/errors/IncorrectAnonymousArgument.md
@@ -5,31 +5,54 @@
 ```purescript
 module Example where
 
-add = (_ + _)
+sub3 = _ - 3
 
-mapArray = map _ [1, 2, 3]
+square = (_ * _)
+result1 = map (_ * _) [1, 2, 3]
+
+result2 = map (div _ 2) [1, 2, 3]
+
+result3 = map (1 + _ / 2) [1, 2, 3]
 ```
 
 ## Cause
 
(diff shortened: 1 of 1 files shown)


errors/IncorrectAnonymousArgument.md
```diff
@@ -5,31 +5,54 @@
 ```purescript
 module Example where
 
-add = (_ + _)
```
milesfrain, line 8: This `add` example is problematic because:
- Not ideal to overwrite Prelude functions, while also using the corresponding infix operator.
- Even with a different name, it's trivially replaced with `add` or `(+)`.
- Unclear whether the intention in the broken case is to reuse the same anonymous arg, or consider these separate args. And if separate args, is the expectation that they are populated in order?

errors/IncorrectAnonymousArgument.md
```diff
@@ -5,31 +5,54 @@
...
+sub3 = _ - 3
 
-mapArray = map _ [1, 2, 3]
```
milesfrain, line 10: I think this second-order example adds another layer of potential confusion for beginners. I changed all these `map` examples to use first-order arguments for `_`.

JordanMartinez:
This PR hasn't been merged because I'm not sure if the new content is actually correct. There's some release notes in PureScript that clarify this further.

PR #372, opened 2021-02-05
Add documentation for escape sequences

stengerh:
This pull requests adds a description of valid escape sequences in string literals. I also briefly describes the differences to Haskell regarding string and chars.

I am still new to PureScript. I discovered this information through experiments and by reading the source code of the PureScript compiler. The proposed changes represent what I wished would have been in the documentation when I started with PureScript. Please let me know if the content or style needs changes.

Thanks!

### DIFF
--- a/language/Differences-from-Haskell.md
+++ b/language/Differences-from-Haskell.md
@@ -52,6 +52,14 @@ ap :: forall m a b. (Monad m) => m (a -> b) -> m a -> m b
 
 There is a native `Number` type which represents JavaScript's standard IEEE 754 float and an `Int` which is restricted to the range of 32-bit integers. In JavaScript, the `Int` values and operations are generated with a `|0` postfix to achieve this, e.g. if you have variables `x`, `y`, and `z` of type `Int`, then the PureScript expression `(x + y) * z` would compile to `((x + y)|0 * z)|0`.
 
+#### Strings
+
+There is a native `String` type which is distinct from `Array Char`. A `String` consists of UTF-16 code units and may contain unpaired surrogates. Working with Unicode code points is supported by library functions. The set of supported escape sequences for string literals is different from both Haskell and JavaScript. 
+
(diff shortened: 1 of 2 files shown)


language/Differences-from-Haskell.md
```diff
@@ -52,6 +52,14 @@ ap :: forall m a b. (Monad m) => m (a -> b) -> m a -> m b
 
 There is a native `Number` type which represents JavaScript's standard IEEE 754 float and an `Int` which is restricted to the range of 32-bit integers. In JavaScript, the `Int` values and operations are generated with a `|0` postfix to achieve this, e.g. if you have variables `x`, `y`, and `z` of type `Int`, then the PureScript expression `(x + y) * z` would compile to `((x + y)|0 * z)|0`.
 
+#### Strings
+
+There is a native `String` type which is distinct from `Array Char`. A `String` consists of UTF-16 code units and may contain unpaired surrogates. Working with Unicode code points is supported by library functions. The set of supported escape sequences for string literals is different from both Haskell and JavaScript.
```
hdgarrood, line 57:
```suggestion
There is a native `String` type which is distinct from `Array Char`. A `String` consists of UTF-16 code units and may contain unpaired surrogates. Working with Unicode code points is supported by library functions. The set of supported escape sequences for string literals is different from both Haskell and JavaScript, and is documented in [Syntax.md](../Syntax.md#escape-sequences).
```

hdgarrood (commented):
Looks good, thank you! Not sure if that link will work.

language/Differences-from-Haskell.md
JordanMartinez: There's no such `escape-sequences` section anymore.

hdgarrood: My point was that this PR introduces an escape sequences section in that file, and that I think we should link to it from here.

PR #371, opened 2021-01-28
Document roles, role inference and role annotations

kl0tl:

PR #373, opened 2021-02-09
Add PureScript 0.14 migration guide

thomashoneyman:
This is a first draft of a migration guide for the 0.14 compiler and core libraries. I believe it covers the most disruptive changes we've made.

If anyone on the core team would like to apply a change to this file that they feel is uncontroversial, please go ahead! You don't need to wait on me to see and apply the feedback. For example, I've included issue and pull request links where possible, but I've missed some -- especially in the library restructuring section as there were so many fragmented ones.

**Notes:**

I haven't included links to the PureScript 0.14 release notes or announcement because they aren't ready yet, but I can update those links at any time. (We could make an issue to circle back to it, too, so that this can be merged).

I don't know where the proper place to put this file is; ideally we'd end up with one of these migration guides for every ecosystem change. I may have missed previous ones we've put together -- if anyone knows the right place for this file please let me know. Otherwise perhaps we could create a `migration-guides` directory?

A huge thank you to @JordanMartinez for his draft of ecosystem updates he's been maintaining since we started this whole release, and to the detailed changelogs, notes, and discussions that @kl0tl, @rhendric, @hdgarrood, @garyb, and others have kept when discussing and implementing these changes.


ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
++ class ListToRow (list :: RowList) (row :: Row Type)
+```
+
+Finally, all type-level declarations are now generalized.
+
+```purs
+data Proxy a
+```
+
+This previously had the `Type`-defaulted kind `Type -> Type`. Now this will be generalized to `forall k. k -> Type`. It is better to be explicit about polymorphism by writing signatures, and the compiler will warn about missing kind signatures when it infers polymorphic kinds.
```
JordanMartinez, line 56: Perhaps the first sentence should say, 
> Previously, the kind for `a` above would implicitly default to `Type` (i.e. `data Proxy (a :: Type)`.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
+
+```purs
+foreign import data Effect :: Type -> Type
+
+-- this is a role annotation, setting Effect's type parameter to be
+-- representational instead of nominal (its inferred role).
+type role Effect representational
+```
+
+In contrast, it is unsafe to allow coercions between `Map k1 a` and `Map k2 a`, even when `Coercible k1 k2` holds. We can annotate `Map`'s key type parameter with a nominal role to prevent coercions:
```
JordanMartinez, line 90: Two things. First, perhaps we should explain why this is unsafe? They can have two different `Ord` instances.

Second, it might help to clarify that it still is safe to coerce the values part of the Map.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
+
+#### The `generics-rep` library has been merged into `prelude`
+
+`Generic` has been moved to the bottom of the core ecosystem because it's widely-used class with compiler support, but it didn't previously cover some common types like `Tuple` or `Either`. This change enables all of the data types across the core libraries which have a sensible `Generic` instance to have one.
+
+**To fix:**
+
+You should remove `generics-rep` as a dependency of your library or application, as most of the relevant code and instances live in the `prelude` package. Code related to `Enum` has been moved to the `enums` package, but you should already depend on this package if you were using it.
+
+You should rename any modules that used `Generic` instances for common type classes, as those modules have been renamed.
```
JordanMartinez, line 177: Perhaps also say here that we recommend following this module naming convention when providing a Generic type class? 
`<full module path to module containing class>.Generic`

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
+
+A number of dependencies have been rearranged in the core libraries so that more types can be reused among them (especially newtypes between `profunctor` and `bifunctors`) and so that fundamental data types like `either` and `tuple` are at the bottom of the dependency hierarchy.
+
+Most of these changes involve instances and should be invisible to user code. However, some functions and modules have changed locations and will require changes in your code if you use them.
+
+**To fix:**
+
+The `lookup` function has moved from `Data.Tuple` to `Data.Foldable` and the import should be updated accordingly:
+
+```purs
```
JordanMartinez, line 270: Shouldn't this be `diff`?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
+
+The `Data.Profunctor.Wrap` and `Data.Bifunctor.Wrap` types have been deleted, as any profunctor or bifunctor should also have a functor instance.
+
+This was the old dependency graph:
+
+![old-dependency-graph](https://user-images.githubusercontent.com/1570964/106351480-6e55a100-62aa-11eb-8e8a-72cb33e354dc.png)
+
+This is the new dependency graph, where the dotted line represents three packages which may be merged together in the future:
+
+![new-dependency-graph](https://user-images.githubusercontent.com/1570964/106351479-6ac21a00-62aa-11eb-8234-803f7868cc53.png)
```
JordanMartinez, line 293: I think these graphs should be shown first in this section.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
+
+Several modules have moved or been renamed, all from the `profunctors` and `bifunctors` packages into the `functors` package. You'll need to update your imports accordingly:
+
+```
+Data.Profunctor.Clown  -> Data.Functor.Clown
+Data.Profunctor.Costar -> Data.Functor.Costar
+Data.Profunctor.Cowrap -> Data.Functor.Flip
+Data.Profunctor.Joker  -> Data.Functor.Joker
+Data.Bifunctor.Product -> Data.Functor.Product2
+```
```
JordanMartinez, line 283: We'd also need to add
```
Data.Bifunctor.Clown -> Data.Functor.Clown
Data.Bifunctor.Joker -> Data.Functor.Joker
Data.Bifunctor.Flip  -> Data.Functor.Flip
```

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,618 @@
...
+
+```purs
+x = unsafePartial fromLeft $ Left 10
+y = unsafePartial fromRight $ Right "hi"
+```
+
+This can be replaced with the new lazy `fromLeft'` and `fromRight'` combined with `unsafeCrashWith`:
+
+```purs
+x = fromLeft' (\_ -> unsafeCrashWith "Unexpected Left") $ Left 10
```
JordanMartinez, line 527:
```diff
- x = fromLeft' (\_ -> unsafeCrashWith "Unexpected Left") $ Left 10
+ x = fromLeft' (\_ -> unsafeCrashWith "Unexpected Right") $ Left 10
```

JordanMartinez (commented):
Very well done! Just a few comments and other suggestions. I'll fix some of the obvious things and leave the rest up for more discussion.

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: Whoops! Yes.

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: Oh, thank you.

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: I'm finding it difficult to strike a balance in this section between a) here's just what you need to do to your code and b) here's what the new feature is all about. I don't want to give a tutorial, but as a new feature it does feel like there's quite a bit of requisite knowledge.

I think I ought to add that unless you know you want polymorphic kinds you should probably add a kind signature fixing everything to `Type`, and if you want something more sophisticated you should consult the release notes and the paper to write the annotation.

JordanMartinez: > unless you know you want polymorphic kinds you should probably add a kind signature fixing everything to Type, and if you want something more sophisticated you should consult the release notes and the paper to write the annotation.

That sounds sensible.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+
+-- now these are equivalent, as expected
+singleton k a <> singleton k b == singleton k (a <> b)
+```
+
+Code using these instances will now append in the correct order.
+
+**To fix:**
+
+Most users should not change their code and will now be using the correct instance. However, if you were intentionally relying on the behavior of this instance then you can recover it by writing a newtype around `Object` and implementing this newtype instance as desired. For example:
```
kl0tl, line 444: Perhaps we can mention `Last` again here, as we do already in the `Map` section?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+```purs
+{ x :: Int, y :: Int, z :: Boolean }
+```
+
+**To fix:**
+
+If you would like to use the new behavior, then your code doesn't need to change. Otherwise, you can use the new `Record.Builder.flip` function to flip the arguments and recover the old behavior.
+
+```purs
+Builder.build (Builder.flip Builder.merge ...)
```
kl0tl, line 510: I think we can reuse the arguments of the previous example, and add a type annotation:

```suggestion
Builder.build (Builder.flip Builder.merge { x: 1, y: "y" }) { y: 2, z: true }
  :: { x :: Int, y :: Int, z :: Boolean }
```

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+  arbitrary = elements $ cons' LT [ EQ, GT ]
+```
+
+Or, for a slightly more performant version which constructs the array and converts it without an intermediate `cons`:
+
+```purs
+import Data.Array.NonEmpty (fromArray)
+
+instance arbitraryOrdering :: Arbitrary Ordering where
+  arbitrary = unsafePartial fromJust $ fromArray [ LT, EQ, GT ]
```
kl0tl, line 571: Don’t we miss a call to `elements` here?

```suggestion
  arbitrary = elements $ unsafePartial fromJust $ fromArray [ LT, EQ, GT ]
```

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
++   (Data.CodePoint.Unicode.codePointFromChar '\n')
+```
+
+## Tooling Changes
+
+In PureScript 0.14, compilation errors are printed to stderr instead of stdout ([#3672](https://github.com/purescript/purescript/issues/3672), [#3839](https://github.com/purescript/purescript/pull/3839)). This is the only change in the compiler that affected tools downstream.
+
+### Spago
+
+Spago does not require changes to be compatible with PureScript 0.14. You will, however, want to update to a [0.14-compatible package set](https://github.com/purescript/package-sets) in your `packages.dhall` file.
```
kl0tl, line 606: Should we link to the initial 0.14 package set release here?

```suggestion
Spago does not require changes to be compatible with PureScript 0.14. You will, however, want to update to a 0.14-compatible package set in your `packages.dhall` file, starting from [psc-0.14.0](https://github.com/purescript/package-sets/releases/tag/psc-0.14.0).
```

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: I'd prefer to write as little as possible here and link to more in-depth documentation like #371. @kl0tl, do you have any suggested phrasing for this section that would help folks get right to any code changes they may need to make and otherwise redirect them to better documentation?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+
+#### The `Newtype` class now has `Coercible` as a superclass and no longer has `wrap` and `unwrap` class members.
+
+Implemented in [purescript-newtype#22](https://github.com/purescript/purescript-newtype/pull/22).
+
+The `Newtype` class exists to support convenient wrapping and unwrapping and to implement the combinators in the `Data.Newtype` module. The `Newtype` class now has `Coercible` as a superclass, which allows most combinators to now be implemented with `coerce` from the new `safe-coerce` library. This makes these combinators zero-cost for a performance gain.
+
+The `Newtype` class no longer has `wrap` and `unwrap` members. Instead, `wrap` and `unwrap` are synonyms for `coerce` and are exported as functions from the module. Most code does not need to change, but if you manually defined a `Newtype` instance in your code it will no longer compile.
+
+In addition, newtypes that do not have their constructors exported can no longer be unwrapped via the `Newtype` class. Most code will be unaffected, as giving a `Newtype` instance to a newtype with hidden constructors was already considered bad practice and usually accidental.
```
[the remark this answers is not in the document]
thomashoneyman: Yes, that's a good idea!

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: Yes! Good catch.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+Data.Generic.Rep.Enum           -> Data.Enum.Generic (in the `enums` package)
+Data.Generic.Rep.Eq             -> Data.Eq.Generic
+Data.Generic.Rep.HeytingAlgebra -> Data.HeytingAlgebra.Generic
+Data.Generic.Rep.Monoid         -> Data.Monoid.Generic
+Data.Generic.Rep.Ord            -> Data.Ord.Generic
+Data.Generic.Rep.Ring           -> Data.Ring.Generic
+Data.Generic.Rep.Semigroup      -> Data.Semigroup.Generic
+Data.Generic.Rep.Semiring       -> Data.Semiring.Generic
+Data.Generic.Rep.Show           -> Data.Show.Generic
+```
```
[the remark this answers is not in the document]
thomashoneyman: It's just that the diff would double the line count here; we could do it, though.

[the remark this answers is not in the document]
rhendric: I think the tabular layout is clearer.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+data Boolean
+foreign import data True :: Boolean
+foreign import data False :: Boolean
+```
+
+The old unary `# Type` syntax for row kinds is deprecated, and it should instead be written as `Row Type`.
+
+```diff
+- class ListToRow (list :: RowList) (row :: # Type)
++ class ListToRow (list :: RowList) (row :: Row Type)
```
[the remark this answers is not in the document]
hdgarrood: I think it would be good to explicitly say that the kind of `RowList` has changed from `Type` to `Type -> Type`, where the argument is the kind of the labels in the row list, so if you had  `RowList` in 0.13.x code, that should now be written as `RowList Type`.

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: I actually think most - maybe even _all_ - of the discussion of roles could be struck from this document. My impression is that most users shouldn't need to care about roles, since role inference should cover the vast majority of cases. Also, since they're new, I think they're unlikely to break any existing code: it's only if you want to _start_ using Coercible that you might need to care, but this document is really aimed at people wanting to get their existing code compiling on 0.14.0, right?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+type role Map nominal representational
+```
+
+### Other Changes
+
+Primes are no longer allowed in JavaScript FFI definitions because PureScript 0.15 will generate ES modules.
+
+**To fix:**
+
+Rename any functions that contain a prime and are implemented via the FFI:
```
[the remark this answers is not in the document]
hdgarrood: I think that's a good idea. The compiler should warn you to do it anyway, but it's probably nicer to say it up front.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+
+```js
+// before
+exports["functionName'"] = function (a) { return a; }
+
+// after: these are common options
+exports.functionNameImpl = function (a) { return a; }
+exports._functionName = function (a) { return a; }
+exports.functionNamePrime = function (a) { return a; }
```
hdgarrood, line 115: I find the "Prime" suffix a bit odd, personally. Have we seen this in the wild?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+#### The `globals` library has been deprecated
+
+The `globals` library provided various functions and values that are in the global scope in JavaScript. The library has been deprecated in particular because PureScript supports multiple backends besides JavaScript, and the relevant functions have been moved or reimplemented in other libraries.
+
+Specifically:
+
+* Number values and operations like `isFinite`, parsing, and formatting have been moved to the `numbers` package, which itself has moved into the core organization.
+* Integer parsing has been removed, as it is already included in the `integers` package.
+* Encoding and decoding URIs and URI components using JavaScript's builtin `encodeURIComponent` and `decodeURIComponent` have been moved to the new `js-uri` package.
+* Unsafe global functions have been removed altogether.
```
hdgarrood, line 209: Can you remind me which ones these were? Was it just the unsafe variants of the URI encoding ones?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+
+```diff
+- import Global (readInt)
++ import Data.Int (radix, fromStringAs)
+
+  x :: String -> Maybe Int
+- x = readInt 2
++ x = fromStringAs (unsafeRadix 2)
++   where
++   unsafeRadix n = unsafePartial fromJust $ radix n
```
hdgarrood, line 245: Sorry to go off topic, but I think we should provide a Radix constructor which always succeeds by clamping the input to the valid range, since this is kind of gross. The `numbers` library already clamps in the `exponential`, `fixed` etc formatters.

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: I actually don't think we should show people these dependency graphs; the vast majority of users shouldn't need to care about them, I think.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+Most `Newtype` instances are derived and will continue to work. If you manually defined a `Newtype` instance for a type, however, you can simply remove the `wrap` and `unwrap` definitions:
+
+```diff
++ instance newtypeAdditive :: Newtype (Additive a) a
+- instance newtypeAdditive :: Newtype (Additive a) a where
+-   wrap = Additive
+-   unwrap (Additive a) = a
+```
+
+Most newtypes either export their constructors and have a `Newtype` instance or hide their constructors and do not have the instance. However, if you have a newtype in your code which has hidden constructors and a `Newtype` instance then you will need to either:
```
hdgarrood, line 323: Will you _need_ to do either of these things? I thought this was just a warning for now. Or are we assuming that the reader wants to fix all errors and warnings?

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+**To fix:**
+
+If you are using the `guard` function, update your imports:
+
+```diff
+- import Control.MonadZero (guard)
++ import Control.Alternative (guard)
+```
+
+If you are using the `MonadZero m` constraint, replace it with `Monad m` and `Alternative m`:
```
hdgarrood, line 347: I think we should also say "if you're providing a MonadZero instance, consider removing it, as the class will be removed in a future release" here.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
+```
+
+Code using these instances will now append in the correct order.
+
+**To fix:**
+
+Most users should not change their code and will now be using the correct instance. However, if you were intentionally relying on the behavior of this instance then you can recover it by writing a newtype around `Object` and implementing this newtype instance as desired. For example:
+
+```purs
+newtype Reversed v = Reversed (Object v)
```
hdgarrood, line 446: I'd suggest using `Object (Dual a)` since we already have a `Dual` newtype in prelude which does what we want here: https://pursuit.purescript.org/packages/purescript-prelude/4.1.1/docs/Data.Monoid.Dual#t:Dual

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: Ah, it was basically all of them, wasn't it? https://github.com/purescript-deprecated/purescript-globals/commit/2d56edd41da66639f667ad97e4293db75c835883#diff-c475094f3ca830f19ce57d7b2928870987347b3f28bdb5af290c543ebd2d7afc

Perhaps this is slightly clearer then?
```suggestion
* Unsafe global functions (i.e. those which lived in `Global.Unsafe`) have been removed altogether.
```

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: I expect the reader will want to at least know how to fix warnings, even if they don’t strictly have to (yet).

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: Perfectly happy with that myself! This is just needed in the meantime.

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: No, I haven’t. I can remove it.

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: I made https://github.com/purescript/purescript-integers/issues/46

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: Right - so do you think it might be worth replacing "need" with "may want to" or something like that?

thomashoneyman: Yes, I'm happy to rephrase to be clear these are warnings, not errors.

ecosystem/purescript-0.14-migration-guide.md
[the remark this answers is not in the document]
JordanMartinez: I prefer the current layout, too. The diff would be too noisy.

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: Actually come to think of it, the leading underscore is probably not ideal either since it'll silence the unused variable warnings (which I would quite like to try to land soon after 0.14.0). https://github.com/purescript/purescript/pull/3819 Perhaps we should use a trailing underscore instead? Or maybe only suggest `Impl`?

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: I was wondering whether we already had a newtype for this exact purpose.

ecosystem/purescript-0.14-migration-guide.md
hdgarrood: I just realised that Data.Int provides `binary :: Radix`, so this could be more nicely written as `fromStringAs binary`?

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: I've added this change.

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: Just a thought, but would linking to [my 5-minute answer to a question on the matter](https://discourse.purescript.org/t/what-is-a-role/2109/2) with a comment of "If you're curious, see this link" suffice? I'll likely document this feature more in my learning repo once `v0.14.0` comes out.

ecosystem/purescript-0.14-migration-guide.md
```diff
@@ -0,0 +1,622 @@
...
++ class ListToRow (list :: RowList) (row :: Row Type)
+```
+
+Finally, all type-level declarations are now generalized.
+
+```purs
+data Proxy a
+```
+
+Previously, the kind for `a` above would implicitly default to `Type` (i.e. `data Proxy (a :: Type)`. Now this will be generalized to `forall k. k -> Type`. It is better to be explicit about polymorphism by writing signatures, and the compiler will warn about missing kind signatures when it infers polymorphic kinds.
```
rhendric, line 56: Written this way, I think ‘this’ needs to be expanded:
```suggestion
Previously, the kind for `a` above would implicitly default to `Type` (i.e. `data Proxy (a :: Type)`). Now the kind of `Proxy` will be generalized to `forall k. k -> Type`. It is better to be explicit about polymorphism by writing signatures, and the compiler will warn about missing kind signatures when it infers polymorphic kinds.
```

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: I've made this change.

ecosystem/purescript-0.14-migration-guide.md
rhendric: I don't think we need more places to read about roles. There's what will hopefully be the formal documentation in #371; there's the documentation for `Coercible` in purescript/purescript#3999, there's the changelog, there's this migration guide... if we do need to keep reexplaining roles for different audiences in this way, then fine, but let's at least give the references a star topology in terms of where they link to for more information. (Meaning, I think we should link to the document in #371 and no other document.)

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: Upon more reflection, I think the graph summarizes some changes (e.g. instances were moved) far quicker than words do. However, it doesn't summarize the heart of the change faster than words do, which is that we don't have two newtypes anymore (e.g. `Joker`, etc).

So, I yield. Harry's point is better.

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: Good point. I'll make that change, too.

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: Fair point!

ecosystem/purescript-0.14-migration-guide.md
JordanMartinez: Either one seems fine to me, but we should explicitly state the leading underscore is not recommended.

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: Yes, that's better. Just saw that myself, good catch!

hdgarrood:
> Otherwise perhaps we could create a migration-guides directory?

I think a migration-guides directory is a great idea. Incidentally we already have an 0.10 migration guide in the `historical` directory, so maybe that could move there?

ecosystem/purescript-0.14-migration-guide.md
thomashoneyman: I've updated this to just link to the document in #371 (anticipating its new location).

thomashoneyman:
I've moved the file to `migration-guides/0.14-Migration-Guide.md` and can move the 0.10 guide to `migration-guides/0.10-Migration-Guide.md` later on.

thomashoneyman:
The historical guides _could_ be moved, but the 0.10 one is quite skimpy and the 0.5-and-before one is a bit of a grab bag. Given the age of those releases, perhaps we should just begin a new tradition with this guide?

JordanMartinez:
I think all guides, however skimpy or terrible, should go in the migration folder. Otherwise, someone will open an issue and PR fixing that in the future.

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,596 @@
...
+```
+
+Most newtypes either export their constructors and have a `Newtype` instance, or hide their constructors and do not have the instance. A new compiler warning is raised for types with a `Newtype` instance but hidden constructors.
+
+To address the compiler warning, you can either:
+
+1. Export the newtype constructors, or
+1. Remove the `Newtype` instance -- a breaking change, but the correct option if you used hidden constructors to make your newtype a smart constructor.
+
+#### The `MonadZero` class has been deprecated.
```
kl0tl, line 307: Compiling core modules exporting MonadZero instances, such as Data.List or Data.Maybe, will yield deprecation warnings until we remove the instances. Should we mention that those warnings can be hidden by installing `psa` and compiling with `psa --censor-codes=UserDefinedWarning`?

kl0tl (approved):
I fixed a few things and left a last suggestion. Otherwise this looks very good to me!

Regarding the historical directory: I wouldn’t mind moving the 0.10 migration guide to the migration-guides directory introduced in this pull request, even if its quite succinct. On the other hand the pre-0.5 changelog isn’t even an attempt of a migration guide so I don’t think we should move it.

migration-guides/0.14-Migration-Guide.md
JordanMartinez: > Should we mention

Yeah, we should.

hdgarrood:
Let's ignore the issue of the 0.10 migration guide and pre-0.5 release notes for now, there's enough going on in this PR already.

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+```diff
+- class ListToRow (list :: RowList) (row :: # Type)
++ class ListToRow (list :: RowList Type) (row :: Row Type)
+```
+
+Finally, all type-level declarations are now generalized. In the below declaration, the kind for `a` would default to `Type`:
+
+```purs
+-- defaults to `data Proxy (a :: Type)`
+data Proxy a
```
hdgarrood, line 54:
```suggestion
data Proxy a

> :k Proxy
Type -> Type
```

I think this is clearer

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+
+## Library Changes
+
+Several core library changes accompanied the PureScript 0.14 compiler release. This guide covers notable changes to the `purescript` (core), `purescript-contrib`, `purescript-node`, and `purescript-web` organizations.
+
+However, the changes listed below are not exhaustive. As of the PureScript 0.14 release, all libraries in these organizations contain a CHANGELOG.md file which lists any breaking changes in their latest release. You should consult the changelogs for any libraries you depend on in addition to this migration guide.
+
+### Library Restructuring
+
+Several core libraries have been restructured. In almost all cases functionality has been preserved, but some functions, types, and / or instances have been moved to new locations.
```
hdgarrood, line 113: This is not very useful in its current form. We should either say what's actually moved from where to where, or we should just omit this section entirely. I'd prefer the former.

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+
+-- Code written for PureScript 0.14 which will continue to work with
+-- PureScript 0.15.
+getFoo = Record.get (Proxy :: Proxy "foo")
+```
+
+
+#### The `generics-rep` library has been merged into `prelude`
+
+`Generic` has been moved to the bottom of the core ecosystem because it's widely-used class with compiler support, but it didn't previously cover some common types like `Tuple` or `Either`. This change enables all of the data types across the core libraries which have a sensible `Generic` instance to have one.
```
hdgarrood, line 158: It's not really accurate to say this change enables those types to have those instances; we could have added them before, but the instances would have just had to live inside the `generics-rep` library. Also I think we should avoid talking about the "why" in this document.
```suggestion
The `Generic` type class now lives in the `prelude` library. This makes it easier for us to provide `Generic` instances for common types which should have them such as `Tuple` or `Either`.
```

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+
+**To fix:**
+
+You should remove `generics-rep` as a dependency of your library or application, as most of the relevant code and instances live in the `prelude` package. Code related to `Enum` has been moved to the `enums` package, but you should already depend on this package if you were using it. If you provide your own Generic type class, we recommend the following module naming convention: `<full module path to module containing class>.Generic`.
```
hdgarrood, line 162: I think "Generic type class" is a bit unclear

```suggestion
You should remove `generics-rep` as a dependency of your library or application, as most of the relevant code and instances live in the `prelude` package. Code related to `Enum` has been moved to the `enums` package, but you should already depend on this package if you were using it. If you provide deriving via Generic for a type class you've defined, we recommend the following module naming convention: `<full module path to module containing class>.Generic`.
```

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+If you used URI encoding and decoding functions, then you should add a dependency on `js-uri` and update your imports:
+
+```diff
+- import Global (decodeURI, decodeURIComponent, encodeURI, encodeURIComponent)
++ import JSURI (decodeURI, decodeURIComponent, encodeURI, encodeURIComponent)
+```
+
+If you used various unsafe functions exposed by the `globals` package, you can in most cases use functions from the `partials` package (like `unsafePartial`) combined with partial functions like `Data.Maybe.fromJust` to produce an unsafe function from a safe one.
```
hdgarrood, line 240:
```suggestion
If you used any of the unsafe functions exposed by the `globals` package, you can in most cases use functions from the `partials` package (like `unsafePartial`) combined with partial functions like `Data.Maybe.fromJust` to produce an unsafe function from a safe one.
```

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+
+#### The dependency graph for `functors` and related packages have changed in order to unify newtypes among the libraries.
+
+Implemented in [purescript-foldable-traversable#131](https://github.com/purescript/purescript-foldable-traversable/pull/131), [purescript-functors#31](https://github.com/purescript/purescript-functors/pull/31), [purescript-profunctor#41](https://github.com/purescript/purescript-profunctor/pull/41), [purescript-distributive#17](https://github.com/purescript/purescript-distributive/pull/17), [purescript-contravariant#30](https://github.com/purescript/purescript-contravariant/pull/30), [purescript-bifunctors#22](https://github.com/purescript/purescript-bifunctors/pull/22), [purescript-tuples#46](https://github.com/purescript/purescript-tuples/pull/46), [purecsript-identity#26](https://github.com/purescript/purescript-identity/pull/26), [purescript-either#64](https://github.com/purescript/purescript-either/pull/64), and [purecsript-const#19](https://github.com/purescript/purescript-const/pull/19).
```
hdgarrood, line 244: I think it might be better not to link to these PRs, especially as we aren't doing it elsewhere in this document. I'm not really sure how much these add and we don't want people thinking they need to read through these before they attempt to migrate to 0.14.

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+
+#### The dependency graph for `functors` and related packages have changed in order to unify newtypes among the libraries.
+
+Implemented in [purescript-foldable-traversable#131](https://github.com/purescript/purescript-foldable-traversable/pull/131), [purescript-functors#31](https://github.com/purescript/purescript-functors/pull/31), [purescript-profunctor#41](https://github.com/purescript/purescript-profunctor/pull/41), [purescript-distributive#17](https://github.com/purescript/purescript-distributive/pull/17), [purescript-contravariant#30](https://github.com/purescript/purescript-contravariant/pull/30), [purescript-bifunctors#22](https://github.com/purescript/purescript-bifunctors/pull/22), [purescript-tuples#46](https://github.com/purescript/purescript-tuples/pull/46), [purecsript-identity#26](https://github.com/purescript/purescript-identity/pull/26), [purescript-either#64](https://github.com/purescript/purescript-either/pull/64), and [purecsript-const#19](https://github.com/purescript/purescript-const/pull/19). 
+
+Mostly discussed in [purescript-profunctor#23](https://github.com/purescript/purescript-profunctor/issues/23).
```
hdgarrood, line 246: Let's remove this too imo

migration-guides/0.14-Migration-Guide.md
```diff
@@ -0,0 +1,597 @@
...
+If you used URI encoding and decoding functions, then you should add a dependency on `js-uri` and update your imports:
+
+```diff
+- import Global (decodeURI, decodeURIComponent, encodeURI, encodeURIComponent)
++ import JSURI (decodeURI, decodeURIComponent, encodeURI, encodeURIComponent)
+```
+
+If you used various unsafe functions exposed by the `globals` package, you can in most cases use functions from the `partials` package (like `unsafePartial`) combined with partial functions like `Data.Maybe.fromJust` to produce an unsafe function from a safe one.
+
+#### The dependency graph for `functors` and related packages have changed in order to unify newtypes among the libraries.
```
hdgarrood, line 242: This should probably be folded into the earlier section about library restructuring

migration-guides/0.14-Migration-Guide.md
hdgarrood: Oh never mind, I didn't spot that the following headings are subheadings of this one.

migration-guides/0.14-Migration-Guide.md
hdgarrood: Never mind

migration-guides/0.14-Migration-Guide.md
hdgarrood: Never mind

hdgarrood (approved):

migration-guides/0.14-Migration-Guide.md
JordanMartinez: This change makes sense and is a bit more readable because there's less content. So :+1: from me.

PR #382, opened 2021-04-15
Update Types.md

gdennie:
`one` is defined in Prelude and is conflicting with the respecification here in a weird way. Compiles now.

### DIFF
--- a/language/Types.md
+++ b/language/Types.md
@@ -213,7 +213,7 @@ Most types can be inferred (not including Rank N Types and constrained types), b
 
 ```purescript
 -- Defined in Data.Semiring
-one :: forall a. (Semiring a) => a
+-- one :: forall a. (Semiring a) => a
 
 -- one can be an Int, since Int is an instance of Semiring where one = 1
 int1 :: Int


JordanMartinez:
Could you clarify? I'm not fully understanding why this change is being proposed.

milesfrain:
I believe the issue is that you can't just copy and paste the entire block here, since it's redefining `one`, which is imported by Prelude by default.
https://github.com/purescript/documentation/blob/master/language/Types.md#type-annotations

milesfrain:
Actually, it's not even a redefinition error, just an incomplete statement as is:
```purs
spago repl
PSCi, version 0.14.0
Type :? for help

import Prelude

> :pa
… -- Defined in Data.Semiring
… one :: forall a. (Semiring a) => a
… 
… -- one can be an Int, since Int is an instance of Semiring where one = 1
… int1 :: Int
… int1 = one -- same as int1 = 1
… -- Or even a Number, which also provides a Semiring instance where one = 1.0
… number1 = one :: Number -- same as number1 = 1.0
… -- Or its polymorphism can be kept, so it will work with any Semiring
… -- (This is the default if no annotation is given)
… semiring1 :: forall a. Semiring a => a
… semiring1 = one
… -- It can even be constrained by another type class
… equal1 = one :: forall a. Semiring a => Eq a => a
… 
… 
Error found:
in module $PSCI
at :5:1 - 5:12 (line 5, column 1 - line 5, column 12)

  The type declaration for one should be followed by its definition.

See https://github.com/purescript/documentation/blob/master/errors/OrphanTypeDeclaration.md for more information,
or to contribute content related to this error.
```

thomashoneyman (approved):

hdgarrood:
I think it would be better to mark this code block in a way that indicates that we shouldn’t attempt to compile it. How exactly are you attempting to compile it?

hdgarrood:
That is: I don’t think it makes sense to assume that you can take any of these markdown files, put every purescript code block together, and expect it to compile.

PR #388, opened 2021-06-22
FFI.md typo

jamesdbrock:


thomashoneyman:
Thanks!

hdgarrood:
I don’t think this was a typo, actually. I think “a type Frob at kind Type” makes more sense than “a type Frob as kind Type”. If “at” sounds weird, then I think we could rewrite to “a type Frob whose kind is Type” or something.

jamesdbrock:
> I don’t think this was a typo, actually. I think “a type Frob at kind Type” makes more sense than “a type Frob as kind Type”. If “at” sounds weird, then I think we could rewrite to “a type Frob whose kind is Type” or something.

Well then oops. I believe you, but this surprises me. I would say “a value Frob as type Type” and I would say “a type Frob as kind Type.”

I will open a reversion PR, then we can decide whether to merge that.

PR #392, opened 2021-06-30
Describe ambiguous matches in instance chains

rhendric:
This documents the new-ish behavior in purescript/purescript#4064; it should be merged only once a release is made which includes that PR (it's about to be reverted at the moment because it's a breaking change).

### DIFF
--- a/language/Type-Classes.md
+++ b/language/Type-Classes.md
@@ -53,6 +53,26 @@ main = do
   log $ myShow MysteryItem -- Invalid
 ```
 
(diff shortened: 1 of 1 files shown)


hdgarrood (approved):
Looks good!

language/Type-Classes.md
```diff
@@ -53,6 +53,26 @@ main = do
   log $ myShow MysteryItem -- Invalid
 ```
 
+When type variables are present in the constraint being solved, the question of whether an instance in a chain matches is a little subtle. A type variable in the constraint is treated existentially—it represents some concrete type, but the constraint solver isn't allowed to assume that it is any particular type. This means that, when evaluating an instance as a solution for a particular constraint, there are three possible outcomes for the instance: it can match, it can fail to match, or it can be ambiguous. An ambiguous instance is one that could match the constraint only if one or more of the variables in the constraint happened to represent a particular type.
```
milesfrain, line 56: Should spaces surround the hyphen/dash in `existentially — it`?

rhendric: The fixed-width font obscures it, but that's an em dash—like this. I see em dashes set off with spaces in newspapers, but for other writing I don't think spaces are typical.

Is a spaced hyphen generally used in place of an em dash in these docs? I can change it if so; the dash was just a reflexive choice.

JordanMartinez (approved):

PR #393, opened 2021-07-02
Note precedence of infix operators created with backticks

milesfrain:
I initially assumed infix operators created with backticks had the lowest precedence.

I couldn't find any PS documentation about this, but assume we're following the same rules as Haskell.
https://stackoverflow.com/questions/8139066/haskell-infix-function-application-precedence

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -451,6 +451,12 @@ Operator sections also work for functions used this way:
 fooBy2 = (_ `foo` 2)
 ```
 
+Infix operators created using backticks are left associative with the highest precedence by default.
+
+``` purescript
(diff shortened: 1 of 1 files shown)


language/Syntax.md
```diff
@@ -451,6 +451,12 @@ Operator sections also work for functions used this way:
 fooBy2 = (_ `foo` 2)
 ```
 
+Infix operators created using backticks are left associative with the highest precedence by default.
```
natefaubion, line 454: PureScript does not allow you to change the fixity of infix operators, so I don't think we need to say "by default".

natefaubion (approved):

hdgarrood (approved):

PR #417, opened 2022-02-14
Type classes: Omit explicit instance name

gasi:
The compiler change to make instance names optional was introduced in v0.14.2:
https://github.com/purescript/purescript/releases/tag/v0.14.2

### DIFF
--- a/language/Type-Classes.md
+++ b/language/Type-Classes.md
@@ -153,8 +153,8 @@ data MyADT
   | Arbitrary Int
   | Contents Number String
 
-derive instance eqMyADT :: Eq MyADT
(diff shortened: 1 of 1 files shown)


JordanMartinez (approved):
LGTM. However, I'm wondering if we should state that instance names are optional and provide an example.

AFAIK, removing instance names altogether hasn't been discussed and they're still used in core/contrib/web/node libraries. A new user might wonder if they are some advanced syntax or something.

JordanMartinez:
This hasn't been merged because feedback hasn't been addressed. It wouldn't take much to finish it.

PR #421, opened 2022-02-22
First draft of documentation update for v0.15

sigma-andex:


guides/FFI.md
```diff
@@ -85,14 +78,12 @@ The general rule regarding types is that you can enforce as little or as much ty
 In PureScript, JavaScript code is wrapped using a _foreign module_. A foreign module is just a CommonJS module which is associated with a PureScript module. Foreign modules are required to adhere to certain conventions:
 
 - The name of the foreign module must be the same as its companion PureScript module, with its extension changed to `.js`. This associates the foreign module with the PureScript module.
-- All exports must be of the form `exports.name = value;`, specified at the top level.
+- All exports must be of the form `export const name = value;`, specified at the top level.
```
JordanMartinez, line 81: This is only true for values, not necessarily functions (as the example below breaks this rule).

guides/Getting-Started.md
```diff
@@ -331,17 +326,6 @@ This illustrates a few points about the way the PureScript compiler generates Ja
...
-
-Spago can also be used to generate CommonJS modules from PureScript code. This can be useful when using NodeJS, or just when developing a larger project which uses CommonJS modules to break code into smaller components.
-
-To build CommonJS modules, use the `spago build` command:
-
-    $ spago build
-    ...
-    Build succeeded.
-
-The generated modules will be placed in the `output` directory by default. Each PureScript module will be compiled to its own CommonJS module, in its own subdirectory.
```
JordanMartinez, line 344: I feel like this is still important to say but for ES modules now.

JordanMartinez (changes requested):
Overall looks good.

Should we create a tag in this repo of its state before any ES modules work gets merged and then link to that for everyone on pre-`0.15.x`?

sigma-andex:
> Overall looks good.
> 
> Should we create a tag in this repo of its state before any ES modules work gets merged and then link to that for everyone on pre-`0.15.x`?

Sounds good

guides/Getting-Started.md
sigma-andex: Ok will add it again with ES modules then

sigma-andex: I have added it again and updated to es modules. However I think this paragraph needs a bigger rewrite, it doesn't make much sense anymore imho.

JordanMartinez:
Perhaps we shouldn't merge this until we're getting closer to the v0.15.0 release?

Regardless, I've made the `pre-0.15.0-era` tag.

JordanMartinez:
I think we'll do so immediately after the `0.15.0` release is made. See also #392

JordanMartinez:
Hmm... Then again, this is the first draft, meaning there are other changes we might want to make. And I already have the 0.14-era tag made.

sigma-andex:
@JordanMartinez should we merge this now?

JordanMartinez:
@sigma-andex I don't think so. My main reasons against merging this now is because `0.15.0` isn't out yet. If it gets delayed by some other issue, then this repo's documentation will be incorrect, and it's one of the first things to which we direct new users. 

On the other hand, we don't have a branch where we can iterate on the `0.15.0` updates to this repo. So, I think instead a new branch should be created `0.15.0-updates` into which we merge this and #392. Then, we can iterate on that branch and then merge that branch into `master` once `0.15.0` comes out.

Issue #419, opened 2022-02-17
Function or value name characters are only defined via Unicode categories


halloleo:
The characters a function or value name can consist of [are defined via Unicode categories](https://github.com/purescript/documentation/blob/master/language/Syntax.md#function-and-value-names). 

This is absolutely unambiguous, but not very helpful for the common reader. I suggest listing for each character category which characters of the extended ASCII character set (0x00 to 0xFF) fall into the category.

I am happy to create a PR for this. 😄

JordanMartinez:
Perhaps we could link to a page that covers each category in more detail?

halloleo:
Not sure if it needs another _page_. I've created an PR where I just add to Syntax.md a table with the "normal" characters .

PR #422, opened 2022-03-01
Details about Unicode categories of function and value name characters.

halloleo:
An added table as fix for #419.


JordanMartinez (approved):
Works for me.

PR #424, opened 2022-03-15
Add tooling and library updates to 0.15.0 migration guide

JordanMartinez:
This further adds upon the work started in #421

### DIFF
--- a/migration-guides/0.15-Migration-Guide.md
+++ b/migration-guides/0.15-Migration-Guide.md
@@ -174,6 +174,7 @@ https://github.com/lebab/lebab#unsafe-transforms
 
 In general though it works well in most of the cases.
 
+See also the ["Migrate to ES Modules"](https://github.com/JordanMartinez/purescript-ecosystem-update/blob/master/src/bash/lib/migrateFfiToEs6.sh) script used in the ecosystem updates for inspiration.
 
 Another option you can try is [`cjstoesm`](https://github.com/wessberg/cjstoesm).
 
--- a/migration-guides/0.15-Migration-Guide.md
+++ b/migration-guides/0.15-Migration-Guide.md
@@ -1,5 +1,13 @@
 # PureScript 0.15 Migration Guide
 
+## Tooling
+
+- `purescript-psa` does not need to be updated.
+- `spago` needs to be updated to `vT.B.D`.
+- `pulp` needs to be updated to `v16.0.0`.
+- `purs-tidy` needs to be updated to `vT.B.D`.
+- Use the `prepare-0.15` `package-sets` branch for `v0.15.0`-compatible libraries
+
 ## ES modules migration guide 
 
 ### tl;dr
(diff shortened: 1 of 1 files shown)


migration-guides/0.15-Migration-Guide.md
```diff
@@ -1,6 +1,14 @@
 # PureScript 0.15 Migration Guide
 
-## ES modules migration guide 
+## Tooling
+
+- `purescript-psa` does not need to be updated.
+- `spago` needs to be updated to `vT.B.D`.
+- `pulp` needs to be updated to `v16.0.0`.
+- `purs-tidy` needs to be updated to `vT.B.D`.
+- Use the `prepare-0.15` `package-sets` branch for `v0.15.0`-compatible libraries
```
thomashoneyman, line 9: At the time of release we'll also release the package set, so I don't think this line should be included

migration-guides/0.15-Migration-Guide.md
```diff
@@ -257,3 +265,48 @@ Basic usage:
 parcel build index.html --no-source-maps --no-optimize --no-scope-hoist --dist-dir "dist/" # bundle for the browser
 parcel build index.html --no-source-maps --dist-dir "dist/" # minified bundle for the browser
 ```
+
+## Breaking Changes Made in Core Libraries
+
+### Changes affecting multiple libraries
+
+- Migrated all FFI to ES modules and dropped support for CommonJS modules.
+- Removed all kind-specific Proxy types (e.g. `SProxy`, `Proxy2`, `Proxy3`, `RLProxy`, etc.)
```
thomashoneyman, line 274: Maybe note "If you are using any of these types, replace them with `Proxy` instead."

migration-guides/0.15-Migration-Guide.md
```diff
@@ -257,3 +265,48 @@ Basic usage:
...
+- Removed `MonadZero` type class and all of its deprecated instances.
+
+### `purescript-prelude` changes
+
+- The data type, `NoConstructors`, often used in `Generic`-related code, was changed to newtype `Void`, enabling one to unwrap the newtype and use `absurd`.
```
thomashoneyman, line 279: If this is non-breaking then I'm not sure it needs to be included; if it _is_ breaking then we can keep it.

JordanMartinez: I'm not sure it's breaking, but it's been labeled as such and does actually enable type class instances to exist that weren't possible before. So, I do think it deserves a mention because many do rely upon `Generic`. That being said, I think the main audience would be people using `Generic` to define things like `GenericShow` type classes.

thomashoneyman (approved):

JordanMartinez:
@thomashoneyman Did you see the last two commits I pushed? I think you approved them around the time I pushed them.

thomashoneyman:
Yep, I think they're OK. I think we should pull your update script out into a standalone gist and link to the gist rather than to the overall repo for conciseness' sake, but for the time being this is a draft and it's fine where it is.

PR #425, opened 2022-03-25
replace ref to deprecated generics-rep

afcondon:
What's in this PR:

changes to (1) "Differences from Haskell" page and (2) Guide page on "Type Class Deriving"  and (3) the "Language: Type Classes" page, that had outdated references to the deprecated `purescript-generics-rep` repo,

Since there is a good example in the Try PureScript examples, i added a link to that (link is to Try PureScript! itself, i don't think you can link to a specific example because of the structure of the link, will correct if i'm wrong on that)


JordanMartinez (approved):

JordanMartinez:
Thanks!

PR #429, opened 2022-03-30
Document breaking change summary in 0.15.0 migration guide

JordanMartinez:
There's still a few todos before this is finished, but this is good enough for today.

### DIFF
--- a/migration-guides/0.15-Migration-Guide.md
+++ b/migration-guides/0.15-Migration-Guide.md
@@ -459,7 +459,7 @@ case foo of
       - `Text.Parsing.Parser.Language` -> `Parsing.Language`
       - `Text.Parsing.Parser.Pos` -> `Parsing.Pos`
       - `Text.Parsing.Parser.String` -> `Parsing.String`
-      - `Text.Parsing.Parser.String.Basic` -> `Parsing.Basic`
+      - `Text.Parsing.Parser.String.Basic` -> `Parsing.String.Basic`
       - `Text.Parsing.Parser.Token` -> `Parsing.Token`
 
 ### `groupAllBy` replaced equality function with ordering function (`purescript-lists`)
--- a/migration-guides/0.15-Migration-Guide.md
+++ b/migration-guides/0.15-Migration-Guide.md
@@ -593,10 +593,10 @@ Thus, `launchAff_`'s type signature was changed to `Aff Unit -> Effect Unit`. By
 - If the "discarding" behavior is desired, replace `launchAff_ do` with `launchAff_ $ void do`
 - If undesired, now you know where to fix your code
 
-### TODO - Miscellaneous changes in `purescript-css`
+### Miscellaneous breakage in `purescript-css`
 
-- Add support for `calc` expressions (#140 by @nsaunders)
-- Update the box-shadow implementation (#88 by @vyorkin)
+- In [purescript-css#140](https://github.com/purescript-contrib/purescript-css/pull/140), support for `calc` expressions was added, but the change affected the names and number of various constructors. See the PR for more info.
+- In [purescript-css#88](https://github.com/purescript-contrib/purescript-css/pull/88), the `box-shadow` implementation was updated to reflect `Clay`, the original inspiration for the library. See the PR for more info.
 
 ### `Color.Scheme.*` modules were removed (`purescript-css`)
 
--- a/migration-guides/0.15-Migration-Guide.md
+++ b/migration-guides/0.15-Migration-Guide.md
@@ -504,10 +504,6 @@ As a result of this change, some operators often used with `<|>` had to change t
 
 See [Unbiasing the Semigroup instance for Map](https://discourse.purescript.org/t/unbiasing-the-semigroup-instance-for-map/1935) and [purescript/purescript-ordered-collections#38](https://github.com/purescript/purescript-ordered-collections/pull/38) for more context.
 
-### `purescript-foreign-object`'s `Semigroup` instance was changed
-
-This section has yet to be written. Including here because it relates to the Map discussion above.
-
 ### Update `frequency` to use `NonEmptyArray` (`purescript-quickcheck`)
 
 In the `0.14.x` ecosystem update, `oneOf` was changed from using a `NonEmptyList` to using a `NonEmptyArray` due to the better syntax support of arrays. However, `frequency` was not also updated. This change fixes that.
--- a/migration-guides/0.15-Migration-Guide.md
+++ b/migration-guides/0.15-Migration-Guide.md
@@ -429,6 +429,61 @@ case foo of
     - Replace usage of such types with `Type.Proxy (Proxy(..))`.
 - Removed deprecated `MonadZero` type class and all of its deprecated instances.
 
+### Removal of deprecated things and their replacements
+
+**To fix**:
+- purescript-arrays
+    - update `Data.Array.group'` to `Data.Array.groupAll`
+    - update `Data.Array.NonEmpty.group'` to `Data.Array.NonEmpty.groupAll`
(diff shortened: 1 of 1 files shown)


migration-guides/0.15-Migration-Guide.md
```diff
@@ -429,10 +429,68 @@ case foo of
...
+### `groupAllBy` replaced equality function with ordering function (`purescript-lists`)
+
+```diff
+- groupAllBy :: forall a. Ord a => (a -> a -> Boolean)  -> List a -> List (NEL.NonEmptyList a)
++ groupAllBy :: forall a.          (a -> a -> Ordering) -> List a -> List (NEL.NonEmptyList a)
+```
+
+**To fix**:
+- use `compare`
+- use <code>compare `on` someFunction</code>
```
thomashoneyman, line 458:
```suggestion
- use <code>compare `on` comparisonFunction</code>
```

migration-guides/0.15-Migration-Guide.md
```diff
@@ -446,18 +504,169 @@ case foo of
...
+
+`launchAff_ :: forall a. Aff a -> Effect Unit` will _silently_ discard what the `Aff` computation returns. This behavior isn't always desirable, but the developer won't know otherwise.
+
+Thus, `launchAff_`'s type signature was changed to `Aff Unit -> Effect Unit`. By making this change, developers are forced to clarify whether they want this discarding behavior or not. It's a minor annoyance (and breaking change) for most developers but otherwise makes the resulting code much safer.
+
+**To fix**:
+- If the "discarding" behavior is desired, replace `launchAff_ do` with `launchAff_ $ void do`
+- If undesired, now you know where to fix your code
+
+### TODO - Miscellaneous changes in `purescript-css`
```
thomashoneyman, line 567: Technically speaking, we only need to include core libraries in the migration guide. We can include libraries outside that if they're particularly widely-used (`aff`, `parsing`), but I don't think we should go much further than that.

migration-guides/0.15-Migration-Guide.md
```diff
@@ -446,18 +504,169 @@ case foo of
...
 
 ## Breaking Changes in the `purescript-web` libraries
 
-This section has yet to be written
+### TODO - `purescript-canvas`
+
+- Support arcs that are drawn counter-clockwise (#58, #83 by @karljs and @JordanMartinez)
+- The `Transform` type now uses the field names `a`, `b`, `c`, `d`, `e` and `f`, instead of `m11`, `m12`, `m21`, `m22`, `m31` and `m32` (#86 by @artemisSystem)
+
+### Arg order for `CSSStyleDeclaration` functions changed (`purescript-cssom`)
```
thomashoneyman, line 606: Related to my earlier comment -- I don't think we should include content for libraries outside of core, except for particularly pervasive libraries if they had a big change (ie. `aff`). If we want to add more content covering non-core libraries, that content should go in the respective organizations (for example, a `purescript-contrib` migration guide) and they can be linked from this migration guide. The migration guide for the compiler & core organization should only include compiler & core.

JordanMartinez: Reasons why I'm against doing this "split the content across the orgs" idea:
- Why split this content into four files that disables a 'find, read, and fix" workflow? Since developers don't necessarily know whether `purescript-foo` is part of the `core` or `contrib` libs (node and web are usually obvious), they have to check two places. Won't these two issues slow down the ecosystem update? Won't that slow down packages from being added to the package set again once they've been fixed? What problem does this solve?
- Where do changelogs for `node` and `web` go since they don't have a governance repo? Is creating a repo for something this small worth it?
- If we're going to be inconsistent (e.g. "We'll only do core libraries and heavily used libraries like `aff`), what counts as 'heavily used? Most code can't be built with "just" core libraries (e.g. Halogen won't compile unless `contrib` libs are also updated, and Halogen is used by a lot of people). Why is one library included but another not?
- Making breaking changes doesn't have a corresponding "guide" (e.g. a release guide in a repo). AFAIK, even writing migration guides in 0.14 was a new thing. I'd argue for making things easier on developers for the time being and figuring out exactly how we want to do breaking changes later.

thomashoneyman: > that disables a 'find, read, and fix" workflow? Since developers don't necessarily know whether purescript-foo is part of the core or contrib libs (node and web are usually obvious), they have to check two places.

That’s a good point. Contrib libraries especially can be confusing to find. I think this is a good reason to keep the content here.

> Why is one library included but another not?

The line I’m trying to tread is this:

1. The core team is responsible for the compiler and core libraries only
2. Some libraries, despite not technically being under our purview, arguably should be included simply because they are basically a core library (e.g. Aff), though maybe this is more an argument to move Aff into core

I’ve also been wanting to keep the signal to noise ratio high in these documents by only including “important” changes (you can see the individual library for all the details). But that’s not so relevant for an aggregated changelog.

———

All this said: I retract, and agree it’s OK to include non-core things in these change summaries. Thanks for helping me see some points I wasn’t considering @JordanMartinez

JordanMartinez: > Some libraries, despite not technically being under our purview, arguably should be included simply because they are basically a core library (e.g. Aff), though maybe this is more an argument to move Aff into core

Doesn't `aff` exist in `contrib` because some backends don't need it whereas the JS backend does? Or put differently, if PureScript formalized a runtime for backends to implement, `aff` could be in core, right?

migration-guides/0.15-Migration-Guide.md
```diff
@@ -446,18 +528,190 @@ case foo of
...
+
+Thus, `launchAff_`'s type signature was changed to `Aff Unit -> Effect Unit`. By making this change, developers are forced to clarify whether they want this discarding behavior or not. It's a minor annoyance (and breaking change) for most developers but otherwise makes the resulting code much safer.
+
+**To fix**:
+- If the "discarding" behavior is desired, replace `launchAff_ do` with `launchAff_ $ void do`
+- If undesired, now you know where to fix your code
+
+### TODO - Miscellaneous changes in `purescript-css`
+
+- Add support for `calc` expressions (#140 by @nsaunders)
```
JordanMartinez, line 593: @nsaunders Would you mind writing instructions for how to migrate 0.14.x to 0.15.x code due to this change?

migration-guides/0.15-Migration-Guide.md
```diff
@@ -450,6 +450,17 @@ case foo of
...
+- purescript-parsing
+  - All modules in this repo starting with `Text.Parsing` or `Text.Parsing.Parser` were renamed to just `Parsing`:
+      - `Text.Parsing.Indent` -> `Parsing.Indent`
+      - `Text.Parsing.Parser` -> `Parsing.Parser`
+      - `Text.Parsing.Parser.Combinators` -> `Parsing.Combinators`
+      - `Text.Parsing.Parser.Expr` -> `Parsing.Expr`
+      - `Text.Parsing.Parser.Language` -> `Parsing.Language`
+      - `Text.Parsing.Parser.Pos` -> `Parsing.Pos`
+      - `Text.Parsing.Parser.String` -> `Parsing.String`
+      - `Text.Parsing.Parser.String.Basic` -> `Parsing.Basic`
```
thomashoneyman, line 462: Should this say Parsing.String.Basic?

JordanMartinez:
@thomashoneyman I've linked to the 2 css PRs and added the parsing module name change. This should be good to go. I can submit another PR with the final ecosystem update once `affjax` PRs get in.

migration-guides/0.15-Migration-Guide.md
JordanMartinez: Fixed.

thomashoneyman (approved):

PR #435, opened 2022-04-23
Add RefexFlags breaking change to strings

JordanMartinez:
:ping_pong: @thomashoneyman


JordanMartinez:
Turns out there's been more changes in `parsing`. Apparently, I'm not watching that repo, but I'm not sure when that change was made. I certainly don't recall doing that intentionally.

Also, I found that all the contrib libraries' PR links are broken. So, I'll just fix this in a new PR.

PR #460, opened 2023-10-29
Update Editor and tool support

postsolar:
- Spago is now the recommended way to bundle JS-backend PS code
- Add `spago-legacy` to the list of deprecated tools
- Replace `purty` with `purescript-tidy` as the recommended formatter; add `purty` to the list of deprecated tools
- Add `purescript-backend-optimizer`
- Add [purerlex](https://github.com/drathier/purerlex) as a Purerl-specific tool

### DIFF
--- a/ecosystem/Editor-and-tool-support.md
+++ b/ecosystem/Editor-and-tool-support.md
@@ -5,9 +5,26 @@ The PureScript ecosystem contains a number of editor plugins and tools to help y
 At minimum, almost all projects will use:
 
 - The [PureScript compiler](https://github.com/purescript/purescript)
-- The [Spago](https://github.com/purescript/spago-legacy) build tool and package manager
+- The [Spago legacy](https://github.com/purescript/spago-legacy) build tool and package manager
+- The [Spago rewrite](https://github.com/purescript/spago), a recent major rewrite of Spago package manager
 
 These tools can be installed via [NPM](https://www.npmjs.com), among other installation methods.
+
+#### Which Spago?
+
+At the moment of writing, Spago is going through a period of transition from an older to a new codebase. Consider the following characteristics of each:
+
+**Spago**
+- The currently maintained and actively developed version of Spago
+- Considered to be in its alpha stage
+
+**Spago legacy**
+- Officially deprecated and not maintained, except for security updates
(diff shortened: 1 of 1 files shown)


ecosystem/Editor-and-tool-support.md
```diff
@@ -74,7 +74,7 @@ There are many tools available to help you develop libraries and applications in
 
 There are several build tools and package managers available for PureScript.
 
-- [spago](https://github.com/purescript/spago) is the standard package manager and build tool for Purescript, powered by [Dhall](https://github.com/dhall-lang/dhall-lang) and [package-sets](https://github.com/purescript/package-sets).
+- [spago](https://github.com/purescript/spago) is the standard package manager and build tool for Purescript.
```
JordanMartinez, line 77: Perhaps this should link to `spago-legacy` because the Spago rewrite is still alpha?

postsolar: I considered this, but I went by the fact `spago-legacy` is officially deprecated and the repo is occupied by the Spago rewrite, which would by extension mean that most of the ecosystem links to the Spago rewrite. Since this page is aimed at first-comers I decided it's worth it to just link to new Spago somewhat ahead of time to avoid extra confusion. If your suggestion is a definitive one I could revert it.

JordanMartinez: I think we should highlight here that the `spago` repo is still in alpha and most learning resources assume `spago-legacy`, even though it's been deprecated. Most people looking at these files will be new learners. So, we don't want to complicate their learning journey by installing the tool most resources don't use.

postsolar: Good points, thanks. I updated the document to link to `spago-legacy` instead.

JordanMartinez: Oh, that's not what I meant. I think both should be linked to in the same section but with clarification on the situation.

postsolar: Oh sorry then, I misunderstood it. I'll link to both and add a note with clarification now.

JordanMartinez (approved):
Nicely said.

PR #462, opened 2024-03-03
Add documentation for qualified operators and type operators

### LINKED ISSUE #4538: Document syntax for qualified operators
## Description

This sounds like a common thing, but I couldn't find an issue about this nor anything in [syntax documentation](https://github.com/purescript/documentation/blob/master/language/Syntax.md#operators-as-values) that would imply it's not allowed intentionally. Given that qualified operators are allowed to be used in the infix form, I don't see why they wouldn't be allowed to be used in their desugared form.

```purescript
module M where

import Data.Either.Nested as E

x = (E.\/)
```

Snippet above currently fails with:
```
[1/1 ErrorParsingModule] src/M.purs:5:6

  13  x = (E.\/)
           ^^^^
```

This is also the case not just with values but also with instance heads, eg `instance C ((M.%) a b)`, and possibly other places.

## To Reproduce

https://try.purescript.org/?code=LYewJgrgNgpgBAWTgdwBYwE4wFDYJbAAOIGALnACICGpVAdAKJ6noZ0ByMAzqTGHFS5wGuAB5wAvHAAUDOgB0A9AEpsQA

## Expected behavior

Code should be parsed and compiled just as with unqualified operators.

## Additional context

I am not affected by this (currently) and I personally don't have a use case for this. I discovered this problem rather incidentally, while toying around.

## PureScript version

0.15.13

postsolar:
Closes https://github.com/purescript/purescript/issues/4538

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -375,14 +375,31 @@ fourToSix = Cons 4 (Cons 5 (Cons 6 Nil))
 oneToSix = oneToThree <> fourToSix
 ```
 
-Operator alias declarations are made up of four parts:
+*Types* can also serve as operators. Consider this example from the module `Data.Tuple.Nested` of `purescript-tuples`:
+```purescript
+infixr 6 type Tuple as /\
+```
+
+Operator alias declarations are made up of five parts:
 
 * The associativity: either `infixl`, `infixr`, or `infix`.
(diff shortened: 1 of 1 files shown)


JordanMartinez (approved):

JordanMartinez:
Thanks!

PR #467, opened 2025-02-01
Add documentation for type holes and language extensions

JoelLefkowitz:
When I came across type holes and applicative do notation for the first time I looked for them in `language/Syntax.md` but couldn't find anything. I think it would be very helpful to include a small section on the syntax and point readers to the `language/Differences-from-Haskell.md` page where the language extensions are described.

### DIFF
--- a/language/Syntax.md
+++ b/language/Syntax.md
@@ -688,7 +688,7 @@ Hole 'x' has the inferred type
 
 ## Language extensions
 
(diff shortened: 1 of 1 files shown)


JoelLefkowitz:
@JordanMartinez could you please take a look at this?

JoelLefkowitz:
@paf31 do you have a moment to please take a look at this?

